PR #3, opened 2026-05-26
Implement B4 gRPC contract (InferenceCache service, fail-open stubs)

EdHasNoLife:
Implements **[CAC-46 / B4](https://linear.app/cachebox/issue/CAC-46/b4-follow-up-define-full-grpc-service-contract-replace-ping)** — replaces the `Ping` placeholder with the full `InferenceCache` gRPC contract from [`docs/design/grpc-contract.md`](docs/design/grpc-contract.md).

## Changes
- **proto** `proto/inferencecache/v1alpha1/inferencecache.proto` — package `inferencecache.v1alpha1`, service `InferenceCache`, the 8 RPCs (`RenderTemplate`, `LookupRoute`, `LookupPDRoute`, `GetCacheState`, `ReportCacheState`, `PublishEvent`, `StreamCacheEvents`, `StreamMetrics`) + messages. Removes the old `cache.v1alpha1` proto + generated code.
- **server** registers `InferenceCache` with **fail-open stub handlers** (`LookupRoute`/`LookupPDRoute`→`NO_HINT`; `RenderTemplate`→`OK`/empty; `ReportCacheState`/`PublishEvent`→`Ack`; event/metric streams close cleanly). Removes the `Ping` placeholder; keeps `grpc.health.v1`.
- **tests** for fail-open `LookupRoute` and `RenderTemplate`.
- Design doc marked implemented (Phase 1 stubs).

Real logic (index-backed lookups, rendering, metrics) lands in later modules. Java stubs come with the gateway client (E1).

> Note: also re-lands the proto↔doc **sync-guard** commit that PR #2's squash dropped (commit `15e8d65`).

## Test plan
- [x] `make pre-pr` clean (ci + no generated-code drift)
- [x] `make build` + `make test` green (incl. new fail-open tests)
- [x] proto↔doc pairing guard satisfied (design doc updated with the proto)
- [x] No `oci`/`oracle` introduced
- [ ] CI green on this PR


EdHasNoLife:
Thanks for the review. Addressed in 87878e3 and clarified below.

### Items to verify

**Error handling in `ReportCacheState`** — good catch; the non-EOF branch was untested. Added `TestReportCacheStateClientCancel` (`pkg/server/server_test.go`): the client aborts mid-stream without half-closing, and we assert the server returns `codes.Canceled` rather than a success `Ack`. That's exactly the *client-disconnect* case — a disconnect surfaces to the handler as a non-EOF `Recv()` error, which the loop propagates (`inferencecache_service.go:44-53`).
- *Channel fills up*: N/A for this handler — there's no internal channel/queue. It's a synchronous drain loop (`Recv` → discard → `Recv`); gRPC flow control applies backpressure on the socket, so there's nothing to overflow. Backpressure/queue tests become relevant in B6 when `ReportCacheState` actually writes to the index.

**Timeout handling** — the `5 * time.Second` is a *test-side call deadline* for the in-process bufconn client, not a server or deployment setting. The server imposes no per-RPC deadline; clients set their own (the contract is fail-open — see `docs/design/grpc-contract.md`, "the server should answer within `slo` / the client's lookup timeout"). Nothing deployment-context-dependent ships here.

**Proto imports** — the full proto source *is* in this PR: `proto/inferencecache/v1alpha1/inferencecache.proto`. It's self-contained — `proto3`, package `inferencecache.v1alpha1`, zero `import` statements (no external/well-known-type deps). The `.pb.go` files are generated from it via `make proto-gen` (drift-checked in CI).

**Service registration** — yes: `New()` registers the descriptor at `pkg/server/server.go:44` (`icpb.RegisterInferenceCacheServer(grpcServer, newInferenceCacheService())`), alongside `grpc.health.v1` at line 43. The new wire-level test reaches the service through a real client stub, so it fails if registration regresses.

### Suggestions

- **Context-cancellation test** — done (see above).
- **Service discovery** — out of scope for this contract PR. Endpoint/discovery (K8s Service + how gateways resolve it) lands with the client-integration work; the contract doc covers the API surface, not deployment topology. Happy to file a follow-up.
- **Backwards compatibility with the HTTP handler** — fully preserved. gRPC and HTTP are independent listeners (`Serve(ctx, grpcListener, httpListener)`); this PR only adds the gRPC service and leaves `/healthz` untouched, as `TestHealthzReturnsOK` confirms. No HTTP client is affected.
- **Buffer size** — `1024 * 1024` is bufconn's in-memory pipe buffer (bytes in flight on the fake socket), unrelated to `CacheStateUpdate` size; it's the standard bufconn default and ample for these metadata-only messages. Added a comment on the helper saying so.

heymrbox (approved):

PR #5, opened 2026-05-27
Add /readyz and Prometheus /metrics to the cache server (B5)

EdHasNoLife:
## Summary
Closes the remaining gaps in **CAC-47 (B5)**. The "register real RPCs / remove the `Ping` placeholder" items already landed in PR #3, so this adds the two outstanding HTTP surfaces on the server's HTTP listener:

- **`/readyz`** — readiness probe. The stub server has no external dependencies to gate on yet, so it tracks liveness; real checks (index warm, backends reachable) land with B6.
- **`/metrics`** — Prometheus endpoint served from a **per-`Service` registry**, emitting `inferencecache_server_up` (tech spec §4.3), set to `1` while `Serve` runs and `0` on shutdown.

## How these endpoints are consumed (roadmap)
These are Kubernetes-facing operational endpoints on the HTTP listener (`:8080`), separate from the gRPC contract (`:9090`):

- **`/readyz` — readiness gate for traffic.** Consumed by the server Pod's **readiness probe** (generated by the controller); Kubernetes only routes Service traffic to a replica once it returns 200. *This PR:* tracks liveness (no deps yet). *Next (B6/CAC-20):* becomes a real **dependency gate** — returns 503 until the CacheIndex has warmed/synced (and later, configured backends are reachable), so the gateway never gets stale `NO_HINT` answers from a server that hasn't loaded state yet. Distinct from `/healthz` (liveness = "restart me").
- **`/metrics` — Prometheus scrape target.** Consumed by Prometheus (ServiceMonitor/PodMonitor) → Grafana dashboards + alerts, and downstream by **F1 (CAC-36)** cache-aware autoscaling. *This PR:* `inferencecache_server_up` plus standard `go_*`/`process_*`. *Next:* the §4.3 schema (`inferencecache_lookup_route_calls_total{reason_code,hint_used}`, `…_latency_seconds`, `hit_rate`, `index_entries`, …) is emitted as handler logic lands (**B6/CAC-20**) and standardized as a published spec (**F3/CAC-38**).

B5 lands the *plumbing* (endpoints + one liveness gauge); the substance arrives with B6 and F3.

## Design notes
- **Per-`Service` registry** (not the global default) keeps the server binary's metrics isolated from the controller binary (which uses controller-runtime's registry) and lets tests construct multiple servers without duplicate-registration panics.
- **Scope:** the fuller §4.3 schema (`hit_rate`, lookup latency, `index_entries`, …) is owned by the standalone metric-schema work (F3/CAC-38). This lands only the endpoint + the documented liveness gauge.
- `/metrics` intentionally also exposes the standard `go_*`/`process_*` collectors (conventional + useful for ops); these are outside the §4.3 schema.
- `prometheus/client_golang` promoted to a direct dependency (it was already transitive — no `go.sum` change needed).
- Consolidated the test server setup into one helper that now also exposes the HTTP base URL; `/healthz`, `/readyz`, and `/metrics` share it.

## Test plan
- [x] `make pre-pr` green (naming + buf lint + fmt + vet + test + build + drift)
- [x] `go test ./pkg/server/... -race` — `/healthz` + `/readyz` return OK; `/metrics` exposes `inferencecache_server_up 1`; existing gRPC/stream tests still pass
- [x] CI green on the PR


EdHasNoLife:
Thanks — all three addressed in aabed91.

- **server_up race (server.go):** moved `s.metrics.up.Set(1)` (with `defer …Set(0)`) to *before* the gRPC/HTTP goroutines launch, so a scrape can never observe `inferencecache_server_up 0` while `Serve` is running. Good catch — this also removes a latent flake in the new test.
- **Go/process collectors vs. "inferencecache_* only" (metrics.go):** kept the standard `go_*`/`process_*` collectors (conventional + useful for ops) and reworded the comment to say `inferencecache_*` is *this project's* schema prefix while the registry additionally exposes the standard collectors, which are outside the §4.3 schema.
- **Flag help (cmd/server/main.go):** updated to "serves /healthz, /readyz, /metrics".

(Re: the sandbox note — tests can't run there because `/tmp` is read-only; they pass locally and in the GitHub `Lint, Test, Proto` job under `-race`.)

heymrbox (approved):

PR #6, opened 2026-05-27
Implement CacheBackend API

heymrbox:
## Summary

Expand the `CacheBackend` v1alpha1 API to match CAC-45 / tech spec §4.1: add deployment, storage, integration, engine selector, backend config, template override, and status health/index fields. Regenerated deepcopy and CRD manifests, updated the sample manifest, and added schema/deepcopy tests for the new validation surface.

## Linear

Closes CAC-45

### Quality
- [ ] CI is green.


EdHasNoLife (commented):
## Automated review — CacheBackend API

Reviewed at extra-high effort (5 finder angles + verifier + sweep). `go build ./...`, `go vet ./...`, and `go test ./...` all pass. The deepcopy functions and CRD schema were checked by inspection against the Go types/markers and look consistent.

**No high-severity correctness bugs.** The remaining items are medium/low design and test-quality observations, posted inline. The two I'd actually act on:

- **`omitempty` on the non-pointer struct fields** (`Storage`, `Integration`, `EngineSelector`) — a no-op in Go, so they always serialize as `{}`. Switch to pointers if absence must be distinguishable (matters for server-side-apply ownership).
- **Endpoint/Health print columns** now read `.status.*`, which the no-op controller never populates — the previously-useful `.spec.endpoint` display goes blank until status reconciliation lands.

**Note (not a finding):** `CacheBackendType` defines six constants but has no `+kubebuilder:validation:Enum` marker, unlike its sibling enum types — so `spec.type` accepts any string. The test `requireNoEnum` asserts this absence is *intentional*; flagging only to confirm that was deliberate.

api/v1alpha1/cachebackend_types.go
```diff
@@ -1,30 +1,175 @@
...
+	DeploymentKind CacheBackendDeploymentKind `json:"deploymentKind,omitempty"`
+
+	// Replicas is the desired number of backend workload replicas.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	Replicas *int32 `json:"replicas,omitempty"`
+
+	// Storage describes persistent storage requested by the backend.
+	// +optional
+	Storage CacheBackendStorageSpec `json:"storage,omitempty"`
```
EdHasNoLife: **`omitempty` on a non-pointer struct field is a no-op.** `Storage` (here), `Integration` (L75), and `EngineSelector` (L79) are value structs, and Go's `encoding/json` never treats a struct value as "empty" — so these always serialize as `storage: {}`, `integration: {}`, `engineSelector: {}` even when the user omitted them.

Under server-side-apply this manager then "owns" those empty objects and can stomp another manager's nested values, and it defeats any future absence-based CRD defaulting. Fix: make them pointers (`*CacheBackendStorageSpec`, etc.) if absence must be distinguishable.

api/v1alpha1/cachebackend_types.go
```diff
@@ -1,30 +1,175 @@
...
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// Health summarizes the observed backend health.
+	// +optional
+	Health CacheBackendHealth `json:"health,omitempty"`
+
+	// IndexEntries is the observed number of cache index entries for this backend.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	IndexEntries int64 `json:"indexEntries,omitempty"`
```
EdHasNoLife: `IndexEntries` is an `int64` with `omitempty`, so the legitimate observed value `0` is dropped from status JSON and is indistinguishable from unset. A backend with genuinely 0 index entries reports no `indexEntries` field at all. Consider dropping `omitempty` or making it a pointer.

config/crd/bases/inferencecache.io_cachebackends.yaml
```diff
@@ -20,7 +20,10 @@ spec:
     - jsonPath: .spec.type
       name: Type
       type: string
-    - jsonPath: .spec.endpoint
+    - jsonPath: .status.health
+      name: Health
+      type: string
+    - jsonPath: .status.endpoint
```
EdHasNoLife, line 26: The `Endpoint` print column was repointed from `.spec.endpoint` to `.status.endpoint` (and a `Health` column added for `.status.health`), but `internal/controller/cachebackend_controller.go` is a no-op that never writes status. Result: `kubectl get cb` shows permanently blank Endpoint/Health columns (not even `Pending`). For an `External` backend the user-supplied `spec.endpoint` used to display here and now doesn't. Fine if status reconciliation lands soon; otherwise it's a UX regression.

api/v1alpha1/cachebackend_types_test.go
```diff
@@ -0,0 +1,256 @@
...
+				Status:             metav1.ConditionTrue,
+				Reason:             "Available",
+				Message:            "backend is ready",
+				LastTransitionTime: metav1.Now(),
+			}},
+		},
+	}
+
+	copied := backend.DeepCopy()
+	backend.Spec.BackendConfig["evictionPolicy"] = "FIFO"
```
EdHasNoLife, line 166: `TestCacheBackendDeepCopyCopiesNestedFields` only mutates and asserts the reference-type fields (the two maps, `Template.Raw`, `Conditions`). It never mutates the pointer/Quantity fields it sets — `Replicas`, `FailOpen`, `StorageClassName`, `LookupTimeoutMs`, `MinimumPrefixTokens`, `PVC.Size`. A future regression that shallow-copied any of those would still pass this test despite the name promising "nested fields".

api/v1alpha1/cachebackend_types_test.go
```diff
@@ -0,0 +1,256 @@
...
+	data, err := os.ReadFile("../../config/crd/bases/inferencecache.io_cachebackends.yaml")
+	if err != nil {
+		t.Fatalf("read generated CRD: %v", err)
+	}
+
+	var crd map[string]any
+	if err := yaml.Unmarshal(data, &crd); err != nil {
+		t.Fatalf("unmarshal generated CRD: %v", err)
+	}
+	return mustPath[map[string]any](t, crd, "spec", "versions", 0, "schema", "openAPIV3Schema")
```
EdHasNoLife: Hardcoding version index `0` is fragile. If a second CRD version (e.g. `v1beta1`) is ever added before `v1alpha1` in the list, index `0` selects the wrong version and the entire field/enum/minimum suite silently validates the wrong schema. Prefer locating the version by `name == "v1alpha1"`.

api/v1alpha1/cachebackend_types_test.go
```diff
@@ -0,0 +1,256 @@
...
+	return mustPath[map[string]any](t, schema, "properties", field)
+}
+
+func requireEnum(t *testing.T, schema map[string]any, expected []string) {
+	t.Helper()
+
+	values := mustPath[[]any](t, schema, "enum")
+	actual := make([]string, 0, len(values))
+	for _, value := range values {
+		actual = append(actual, value.(string))
```
EdHasNoLife: Unchecked type assertion `value.(string)` (the rest of this helper file uses comma-ok). If a regenerated CRD ever emitted a non-string enum value, this panics with a raw runtime panic instead of a readable `t.Fatalf`. Low severity since current enums are all strings.

api/v1alpha1/cachebackend_types_test.go
```diff
@@ -0,0 +1,256 @@
...
+			return
+		}
+	}
+	t.Fatalf("required fields = %v, want %q", values, field)
+}
+
+func requireNotRequired(t *testing.T, schema map[string]any, field string) {
+	t.Helper()
+
+	values, ok := schema["required"].([]any)
```
EdHasNoLife: `requireNotRequired` returns (passes) silently when the `required` key is absent or not a `[]any`. Combined with a typo'd path this degrades to a vacuous pass — asymmetric with `requireRequired`, which `Fatalf`s on a missing key. The negative assertion can give false confidence that it actually verified "field is not required".

config/crd/bases/inferencecache.io_cachebackends.yaml
heymrbox: Yes, status reconciliation lands in near future

EdHasNoLife (approved):

PR #7, opened 2026-05-27
Back LookupRoute with an in-memory CacheIndex (B6 engine)

EdHasNoLife:
## Summary
Implements the **index-engine half of B6 (CAC-20)** — real ingestion + ranked lookups behind the previously fail-open stubs. The `CacheIndex` CRD + controller status surface is split to **CAC-50** (kept this PR server-only and focused).

New package `pkg/index`: a concurrent, soft-state aggregator keyed by `(tenant, model, hash_scheme, prefix_hash)` → replica entries (+ per-replica stats).

- **`ReportCacheState`** ingests authoritative updates — idempotent on `(replica, hash_scheme, prefix_hash)`.
- **`PublishEvent`** applies `PREFIX_EVICTED` / `REPLICA_UPDATED` / `ALL_CLEARED` deltas (`PREFIX_ADDED` refreshes known entries; `ReportCacheState` is authoritative for additions).
- **`LookupRoute`** returns replicas holding the exact prefix *within a matching `hash_scheme`*, ranked by matched-tokens × freshness; no match → empty + `NO_HINT` (fail open, side-effect-free apart from metrics).
- **`GetCacheState`** returns the `(tenant, model)` aggregate (replica stats + prefix count).
- **Soft state:** TTL eviction goroutine (tied to `Serve`'s ctx) + a max-entries cap bound memory. `prefix_hash` stays engine-opaque.
- **Ops:** `/readyz` now consults `index.Ready()`; `/metrics` adds `inferencecache_index_entries{model}`, `inferencecache_lookup_route_calls_total{model,reason_code,hint_used}`, `inferencecache_lookup_route_latency_seconds{model}`. A drained model's `index_entries` series is zeroed rather than left stale.

## Design notes
- `pkg/index` is **decoupled from proto** (pure domain types); the service translates proto ↔ index, so the index stays unit-testable in isolation.
- Engine-warm `/readyz` gating (waiting for initial KV-event sync) arrives with the C1 hook; today the index is ready as soon as `Serve` starts it.
- The full §4.3 metric schema is standardized in F3/CAC-38; this lands the index/lookup subset.

## Request Flow
A new request arrives whose stable prefix hashes (in vLLM's scheme) to X:
  ```
Gateway → LookupRoute{
    model_id: "llama-3-70b", tenant_id: "acme",
    prefix_hash: X, prefix_token_count: 2048, hash_scheme: "vllm"
  }
  InferenceCache → {
    reason_code: "PREFIX_MATCH",
    replica_scores: [
      {replica_id: "replica-1", score: 0.95, matched_tokens: 2048, est_cache_hit_prob: 0.95},
      {replica_id: "replica-2", score: 0.80, matched_tokens: 2048, ...}
    ],
    lookup_latency_us: 90
  }
```

## Test plan
- [x] `make pre-pr` green (naming + buf lint + fmt + vet + test + build + drift)
- [x] `go test ./pkg/... -race` — index unit tests (ranking/top-K, idempotency, tenant + hash-scheme isolation, TTL eviction, max-entries cap, drained-gauge zeroing, ready lifecycle) + over-the-wire `ReportCacheState`→`LookupRoute` `PREFIX_MATCH`
- [x] Live smoke test (`grpcurl`): ingest → `PREFIX_MATCH` (replica ranked, matched_tokens=128), cross-scheme → `NO_HINT`, `GetCacheState` aggregate, `/metrics` shows `index_entries` + `lookup_route_calls_total`
- [ ] CI green on the PR


EdHasNoLife:
Thanks — this is a real contract ambiguity, addressed in 430b1b4 by making the semantics explicit rather than switching behavior. The key decision: **`CacheStateUpdate` is an incremental delta, not a snapshot.**

Why delta (not snapshot-replace):
- The engine integration (C1/CAC-21) is **KV events** — vLLM emits `BlockStored` / **`BlockRemoved`** / `AllBlocksCleared`. Removals are explicit events, not "absent from the next batch."
- A replica can hold 10k–100k+ cached prefixes; resending the full set on every update would be bandwidth-prohibitive and defeats the streaming/event design.
- **Soft state** (tech spec): a stale hint causes a cache *miss*, never a wrong answer, and TTL bounds the staleness. So `p1` lingering until its `PREFIX_EVICTED` event or TTL is intentional and fail-soft, not a correctness bug.

So the defect you caught was the **wording**: the code said "authoritative" (reads as snapshot) while behaving as additive/refresh. Fixed by nailing the contract in three places:
- `pkg/index/index.go` `Ingest` doc — "additive deltas, NOT snapshots; removals via CacheEvent / TTL."
- `pkg/server/inferencecache_service.go` `ReportCacheState` doc.
- `docs/design/grpc-contract.md` — new **"Additive `CacheStateUpdate`"** contract guarantee.

Removals are already handled: `PublishEvent` `PREFIX_EVICTED` / `ALL_CLEARED` (tested in `TestApplyEventEvictAndClear`) + the TTL eviction sweep. No snapshot-replace, by design.

EdHasNoLife:
Addressed in c8e5642.

**Should-fix (stats replica id):** fixed. The top-level `CacheStateUpdate.replica_id` is now authoritative — the index overrides `ReplicaStats.ReplicaID` with the top-level id when storing (so the payload always matches its key, regardless of what an adapter puts in the nested field), and `updateFromProto` reads the top-level id at the translation boundary. Regression test `TestStatsKeyedByTopLevelReplicaID` ingests a deliberately mismatched nested id and asserts `CacheState` reports the top-level one (with the stats payload intact).

**Nit (wrap Recv error):** intentionally leaving it raw. The non-EOF `Recv` error is already a gRPC `status` error (e.g. `codes.Canceled` when a client aborts). Returning it unwrapped preserves that code over the wire — which `TestReportCacheStateClientCancel` asserts (`status.Code(err) == codes.Canceled`). Wrapping with `fmt.Errorf("...: %w", err)` risks the server remapping it to `codes.Unknown`, degrading the status the client sees. Propagating an already-structured status error verbatim is the behavior we want here.

EdHasNoLife:
Both addressed in adcc348.

- **Should-fix (doc out of sync):** updated `docs/design/grpc-contract.md` — the status line now lists B6, and the scope section has a "B6 (cache index)" update noting `LookupRoute` / `ReportCacheState` / `PublishEvent` / `GetCacheState` are index-backed (with `RenderTemplate` / `LookupPDRoute` / streams still fail-open stubs). No longer contradicts the implementation.
- **Nit (`WithSweepInterval(0)` panic):** `index.New` now clamps non-positive TTL and sweep interval to their defaults, so a misconfigured option can't panic `time.NewTicker(0)` (or divide-by-zero in freshness). Covered by `TestNonPositiveDurationsClampToDefaults`.

EdHasNoLife:
All three addressed in 8f99dd0.

- **Blocking (cross-scheme refresh):** fixed. Since `CacheEvent` has no `hash_scheme` and `prefix_hash` is only meaningful within a scheme, events no longer touch scheme-specific prefix freshness. `PREFIX_ADDED` is now a **no-op** (`ReportCacheState` is authoritative for adds/refreshes); `REPLICA_UPDATED` refreshes only the replica's **stats** liveness; `PREFIX_EVICTED` / `ALL_CLEARED` removals stay (removal is conservative — at worst a cache miss, soft state). New test `TestPrefixAddedEventDoesNotRefreshAcrossSchemes` proves a `PREFIX_ADDED` event lets both schemes' entries expire on schedule.
- **Should-fix (index lifecycle):** `Serve` now derives `serveCtx, cancel := context.WithCancel(ctx)` with `defer cancel()` and starts the index on `serveCtx`, so the eviction goroutine and `Ready()` are torn down on **any** return — including the internal error branch — not just when the caller's ctx is done.
- **Nit (doc scope):** tightened `grpc-contract.md` — the "still out of scope" line now scopes to the event/metric **streams** (M10), and the B6 update spells out the scheme-safe event behavior + that `/metrics` carries the index/lookup metrics.

EdHasNoLife:
Addressed in ea216a6.

**Blocking (empty `hash_scheme`):** fixed — the index now fails open on a missing scheme:
- `Ingest` drops prefix entries when `hash_scheme == ""` (they can't be matched safely; stats stay scheme-independent).
- `Lookup` returns no hint when `hash_scheme == ""`, even if a real scoped entry exists.

So a `CacheStateUpdate`/`LookupRouteRequest` that omits the tag can never collapse engines into one domain. Documented under the "Engine-opaque `prefix_hash`" guarantee in `grpc-contract.md`, with `TestEmptyHashSchemeFailsOpen` covering both the ingest-drop and lookup-NO_HINT paths.

EdHasNoLife:
Addressed in ee089a3.

**Should-fix (stale `index_entries` under concurrent reporters):** fixed. `reportEntries` now takes the per-model count snapshot **while holding `reportMu`** instead of before it. Reporters are serialized by `reportMu`, and each snapshot reads live index state at publish time (mutations complete under `i.mu` before `reportEntries` is called), so whichever reporter runs last always writes the current count — no stale earlier snapshot can win. Lock order stays `reportMu → i.mu` (no caller holds `i.mu` when calling `reportEntries`), so no deadlock. Added `TestConcurrentIngestReportsFinalCount` (50 concurrent ingests, asserts the final gauge equals the live count; passes repeatedly under `-race`).

heymrbox (approved):

PR #8, opened 2026-05-27
Harden CI: race tests, govulncheck (advisory), Dependabot

EdHasNoLife:
## Summary
Tier-1 CI/CD hardening (CAC-51), independent of the B6/CAC-50 work.

- **Race detector in CI** — new `make test-race` (`go test -race -count=1 ./...`) runs in CI, and `make ci` now uses it so the **pre-push hook** catches data races too. The cache index is concurrency-heavy, so this guards a real risk class.
- **govulncheck (advisory)** — new `make vulncheck` (pinned `golang.org/x/vuln`) scans deps + reachable code for known Go CVEs. Wired into CI with `continue-on-error: true` for now: it currently reports CVEs in the Go toolchain stdlib + `x/net` + `grpc` that need a deliberate **Go 1.25 / grpc upgrade** (tracked in **CAC-52**). Flip to blocking once that lands.
- **Dependabot** — `.github/dependabot.yml` for `gomod` + `github-actions`, weekly, minor/patch grouped to limit PR noise. (It will also start proposing the `x/net`/`grpc` bumps CAC-52 needs.)

## Why advisory for govulncheck
A blocking gate would require bumping the Go toolchain to 1.25.x and `grpc` 1.68→1.79 (which can ripple into the generated stubs) — out of scope for a CI-config PR and risky to bundle. Landing the scan now gives visibility + the capability; CAC-52 drives the called-vuln count to zero, then we drop `continue-on-error`.

## Test plan
- [x] `make test-race` green locally
- [x] `make pre-pr` green (now race-tests via `make ci`)
- [x] `make vulncheck` runs (reports the known toolchain/dep CVEs → CAC-52)
- [ ] CI green on the PR (vuln step shows advisory warning, does not fail the job)
- [ ] Dependabot picked up after merge


EdHasNoLife:
Addressed in 221f546. Updated `CONTRIBUTING.md` to match the new gate: the baseline checklist now uses `make test-race` (and adds `make vulncheck`), with a note that `make test` is the faster non-race variant; the pre-push `make ci` description now reads "race tests" instead of "test".

EdHasNoLife:
Addressed in 92c5fa8 — `make vulncheck` in the CONTRIBUTING baseline list is now explicitly labeled advisory (reports known CVEs pending the Go/grpc upgrade; CI doesn't fail on it), matching the `continue-on-error` step in ci.yml.

heymrbox (approved):

PR #10, opened 2026-05-27
Add vLLM + LMCache reference-stack manifests, runbook, and KV-event tooling (A2)

EdHasNoLife:
## Summary

Adds a reproducible **vLLM + LMCache reference stack** under `docs/reference-stack/` — a minimal, explicit deployment that serves a model with LMCache as its KV-cache backend and publishes KV-cache events over ZMQ, so cache-aware behaviour (a prefix-cache hit) and the event stream can be tested end to end.

- **`manifests/`** — GPU reference Deployment + Service: vLLM + LMCache (`--kv-transfer-config` `LMCacheConnectorV1`, `--kv-events-config` ZMQ publisher emitting `BlockStored`/`BlockRemoved`/`AllBlocksCleared`). Plus a **CPU-only variant** (no LMCache) that runs without a GPU.
- **`GPU-RUNBOOK.md`** — VRAM sizing math, model→card/tensor-parallel table, host-resource requirements, and an OCI-shape worked example.
- **`helm/values-reference.yaml`** — upstream vLLM Production-Stack chart path (router disabled), as an alternative to the raw manifests.
- **`scripts/`** — ZMQ KV-event subscriber, prefix-cache-hit test, a synthetic publisher, and a redaction self-test.
- **`VERSIONS.md`** — pinned image/model/chart; GPU image ships a non-applyable placeholder digest that must be pinned before deploy.

Also adds a CI/hook gate (`make verify-no-internal-refs`) banning internal issue-tracker references in tracked files, and removes them from the docs/PR template (public repo).

Docs/tooling only; no Go/proto/CRD touched. Vendor-neutral identity intact (OCI references are confined to clearly-labeled GPU-shape examples in `docs/`, never core identity).

## Verification

**CPU path — verified end-to-end** (vLLM `0.21.0` CPU image, arm64, run locally):

- Prefix-cache hit: cold request **31.06s** → same-prefix request **1.42s**; `vllm:prefix_cache_hits` **0 → 2176**.
- KV-cache events: real `BlockStored` events captured over ZMQ (block-hash chains, `block_size 128`).
- Token redaction: the subscriber stripped `token_ids`, keeping only `token_count` — confirmed against real engine output (metadata-only contract holds).
- Note: this corrected an earlier assumption — vLLM's v1 engine **does** run on CPU and **does** publish KV events there (it just needs enough RAM; the CPU runtime baseline is ~5 GiB before the KV cache).

**Also verified:**

- [x] All manifests pass `kubectl apply --dry-run=client`.
- [x] Synthetic-publisher → subscriber round-trip + `test_kv_events.py` (decode + token redaction).
- [x] `make pre-pr` green (naming + internal-refs + fmt + vet + test + build + no drift); CI green.

**Not yet verified — needs a GPU host:**

- [ ] The **LMCache GPU stack** (the `manifests/` GPU Deployment) — real cache-hit + captured ZMQ sample on an NVIDIA GPU. The CPU run exercises the same engine-config/event path without LMCache offload.


EdHasNoLife:
Addressed the Codex review in 8c7064d:

- **Blocking — token content in events:** `kv_events_subscriber.py` now redacts `token_ids` at the decode boundary and emits `token_count` only, so the committed `captures/*.jsonl` carries metadata only (hashes/counts), never prompt-derived token content — matching the `CacheStateUpdate`/`PrefixEntry` metadata-only contract. Typed-decode failures now emit shape only (`bytes`), never the raw payload. The synthetic publisher no longer bakes in `token_ids`. Verified locally: `token_ids` no longer appears in subscriber output.
- **Should-fix — `latest` in the canonical template:** the tag is now explicitly marked a non-canonical placeholder in both `manifests/deployment.yaml` and `helm/values-reference.yaml`, with the digest-pin called out as the single value M5/the fleet substitutes before any GPU run.
- **Should-fix — CacheBackend field mapping:** reframed as the tech-spec §3.1 *target* shape, noting v1alpha1 currently implements only `spec.type` + `spec.endpoint` (per `api/v1alpha1/cachebackend_types.go`), with the richer fields landing via B1/B2.

EdHasNoLife:
Round 2 addressed in d6563ba:

- **Committed bytecode:** removed `scripts/__pycache__/*.pyc` and added `__pycache__/` + `*.pyc` to `.gitignore`.
- **`:latest` reaching the fleet:** the image is now a *non-applyable placeholder* — `deployment.yaml` uses an all-zero `@sha256:` digest and `helm/values-reference.yaml` uses `REPLACE_WITH_PINNED_DIGEST`, so the raw manifest cannot run an unpinned image; a real digest must be substituted first (VERSIONS.md updated to match).
- **Swallowed decode errors:** the `UNDECODED` path now reports the exception class/message (actionable on schema drift), still without ever emitting the raw payload.
- **Reproducible test:** added `scripts/test_kv_events.py` — broker-free, pytest-free round-trip + redaction tests (asserts `token_ids` never surfaces and `token_count` is preserved). 4/4 pass locally.
- **Nit:** dropped the unused `msgpack` dependency from `requirements.txt`.

heymrbox (approved):

PR #15, opened 2026-05-27
Add vLLM KV-event subscriber → ReportCacheState (C1)

EdHasNoLife:
## Summary

Adds the **vLLM KV-event subscriber** — a sidecar that subscribes to a vLLM engine replica's KV-cache events over ZMQ and reports cache state to the policy server, keeping the B6 `CacheIndex` live from real engine events. **Metadata only — never tokens or prompt text.** Lands in `pkg/adapters/engine/` + a new `cmd/kvevent-subscriber` binary.

- **`events.go`** — decodes vLLM's msgpack `EventBatch` wire format (msgspec array-tagged) into `BlockStored` / `BlockRemoved` / `AllBlocksCleared`. `token_ids` / `parent_block_hash` / `lora_id` are **never materialized** into the Go types (metadata-only by construction); unknown event tags are skipped (forward-compatible).
- **`mapper.go`** — translates to the merged gRPC contract: `BlockStored` → `CacheStateUpdate` (additive, via `ReportCacheState`), `BlockRemoved` → `PREFIX_EVICTED`, `AllBlocksCleared` → `ALL_CLEARED` (via `PublishEvent`). uint64 block hashes encode as 8-byte big-endian `prefix_hash`; every message is stamped with replica/model/tenant/`hash_scheme`.
- **`forwarder.go`** — `Reporter` debounces adds over a window onto a long-lived `ReportCacheState` stream; removals go via unary `PublishEvent`. **Fail-soft**: reconnect-with-backoff, errors logged not propagated, a clear supersedes buffered adds, graceful-shutdown final flush.
- **`subscriber.go`** — ZMQ SUB loop (pure-Go `go-zeromq/zmq4`) behind a `frameSource` interface; reconnects with backoff; malformed batches dropped.
- **`cmd/kvevent-subscriber`** — sidecar entrypoint (engine endpoint, topic, server, replica/model/tenant/hash-scheme, window) + signal-driven drain.

No proto/CRD changes. Vendor-neutral; new deps: `vmihailenco/msgpack/v5`, `go-zeromq/zmq4`.

## Verification

**Unit / integration (CI):**
- [x] decode (incl. >2⁶³ hashes, malformed input, unknown-tag skip), mapping, hash encoding, config validation.
- [x] forwarder over `bufconn` to a recording gRPC server (stream adds, `PublishEvent` removals, clear-supersedes-adds).
- [x] subscribe loop via a fake frame source (decode + skip-malformed + reconnect).
- [x] `make pre-pr` green (naming + fmt + vet + test + build + no drift).

**Live end-to-end (run locally, GPU-free):** real vLLM `0.21.0` CPU engine → ZMQ → `kvevent-subscriber` → policy server → B6 index. Two prefix requests populated the index: `inferencecache_index_entries{model="qwen"} 17`. Confirms the pure-Go ZMQ ↔ vLLM libzmq interop and the full add path against the real server.

## Follow-ups (out of scope)

- ZMQ replay (`buffer_steps`/ROUTER) for gap recovery on reconnect — soft state tolerates loss for now.
- A container image + sidecar manifest for the subscriber.


EdHasNoLife:
Addressed the Codex review in the latest commit (also rebased onto current main):

- **Blocking — hash type:** block hashes are now opaque `[][]byte`. The decoder accepts both vLLM `ExternalBlockHash` variants — binary passes through verbatim, integer hashes normalize to 8-byte big-endian. Previously only `uint64` was decoded, so a byte-hash engine's batches would have been dropped as undecodable. Added a byte-hash wire test (`TestDecodeByteHashes`).
- **Should-fix — graceful drain:** the reporter now runs on a background context and stops by draining a *closed* `out` channel (the subscriber owns the signal context), so batches already buffered at shutdown are flushed instead of lost to a ctx-cancel race. The final flush also drops and reopens the report stream under a detached context (the run-ctx stream is dead after cancel). Added `TestReporterFlushesPendingOnShutdown`.
- **Nit — `-window=0` panic:** `NewReporter` clamps a non-positive window (and backoff). Added `TestNewReporterClampsWindow`.

Verified locally: `make verify-no-internal-refs`, race tests, and the coverage gate (78.6% ≥ 65%) all pass.

EdHasNoLife:
Round 2 addressed in the latest commit:

- **Bounded gRPC contexts:** every RPC now uses its own time-bounded context — each flush opens a fresh, bounded `ReportCacheState` stream (send + `CloseAndRecv`), and `PublishEvent` is per-call bounded (default 5s, `WithRPCTimeout`). A stalled/unreachable server can no longer block the loop indefinitely, restoring the fail-soft contract. This also removed the long-lived-stream lifecycle.
- **Malformed known tags rejected:** a `BlockStored` truncated before `block_size` (or any known tag with too few fields) is now an error and the batch is dropped/logged, instead of being indexed with `token_count=0`. Added `TestDecodeBlockStoredMissingBlockSizeIsError`.
- **`pendTs` reset** after each flush, preserving "0 = server now" for a later non-timestamped batch.
- **Subscriber backoff clamp:** `NewSubscriber` clamps a non-positive backoff, matching `Reporter`.

Verified: race tests clean, coverage 79.0% ≥ 65%, `make pre-pr` green.

EdHasNoLife:
Round 3 addressed in b838695: `StoredPrefixes` now reports a **cumulative** token_count per block hash — `(i+1)*block_size` — since vLLM block hashes chain their parent (block i covers i+1 blocks), so the matched_tokens ranking signal is preserved instead of flattened. It counts within-event tokens only (the parent prefix's length isn't in the event) and never touches token contents; mapper test asserts [128, 256]. Also documented that the EventBatch may carry trailing fields (e.g. a data-parallel rank) which the decoder ignores. Race + coverage (78.8%) + pre-pr green.

EdHasNoLife:
Round 4 (blocking) addressed in 5126603: the `BlockRemoved` path now **flushes pending adds before publishing the eviction**, so a block stored-then-evicted within one debounce window reaches the server in store→evict order — the additive add can no longer overtake the removal and re-create the evicted prefix. Added an over-the-wire test (`TestReporterFlushesAddsBeforeRemoval`) asserting the add is received before the evict. Race + coverage (79.1%) + pre-pr green.

EdHasNoLife:
Addressed the latest should-fix in aac4782: a `BlockStored` with `block_size <= 0` is now rejected during decode (fails open) rather than forwarding a zero/negative `token_count` that could yield a bogus `PREFIX_MATCH` hint. Added a test covering 0 and -1. Race + coverage + pre-pr green; no blocking items outstanding.

EdHasNoLife:
Addressed the should-fix in a686dd4: `dialZMQ` now closes the SUB socket on the `Dial` failure path too (it only closed on the subscribe-failure path), so a bad/unavailable endpoint no longer leaks a socket every reconnect-backoff cycle. Also synced the branch with main (resolved a go.sum conflict via `go mod tidy`); CI + pre-pr green.

EdHasNoLife:
Addressed in 0cdfad2: `hashToBytes` now guards `v == nil` and returns an error, so a msgpack nil in `block_hashes` is logged + skipped (fail-soft) rather than reaching the reflect path. Added a regression test (`[]interface{}{nil}` in block_hashes).

heymrbox (approved):

PR #16, opened 2026-05-27
Add CacheIndex CRD + controller status surface (B6 status half)

EdHasNoLife:
## Summary
The **status-surface half of B6 (CAC-50)** — exposes the server's in-memory cache aggregate as a cluster-scoped, status-only `CacheIndex` CR so operators can `kubectl get cacheindex`. (Engine half landed in CAC-20/#7.)

Design (per the locked decision): **server exposes the data, controller writes the CR** — no proto/contract change, reuses the controller's existing k8s client + RBAC.

- **CRD** (`api/v1alpha1/cacheindex_types.go`) — cluster-scoped, status-only. Status mirrors the tech spec: `replicas[]{id,cacheMemoryBytes,hitRate,pressure,lastUpdate}`, `prefixes.summary{total,hot=0}`, `tenants[]{id,memoryUsed,hitRate}`. Rates are **decimal strings** (`"0.78"`) — CRDs avoid floats for cross-language portability (controller-gen rejects floats without `allowDangerousTypes`, and a Java client is coming). `hot=0` until per-prefix access-counting exists.
- **`pkg/index` `Snapshot()`** — cluster-wide aggregate: latest stats per replica id; tenant memory/hit-rate dedup replicas within a tenant (documented approximation).
- **`pkg/server` `/snapshot`** — internal HTTP JSON endpoint (alongside `/healthz`,`/readyz`,`/metrics`). **Metadata only** — replica/tenant stats + prefix counts, never KV tensors or prompt text.
- **`internal/controller` `CacheIndexPoller`** — leader-elected manager `Runnable` (not an event-driven reconciler — the data source is the server, not the CR). Maintains the singleton `cluster-default` CR; writes status only when the meaningful aggregate changed (timestamps ignored for change detection → no churn under steady traffic).
- **`cmd/controller`** — wired with `--server-snapshot-url` (default `http://inference-cache-server:8080/snapshot`) and `--cacheindex-refresh-interval` (default 30s).

## Notes / scope
- **Single server instance assumed** (Phase-1 in-memory index); HA cross-instance aggregation is later.
- `/snapshot` is unauthenticated on the internal `:8080` HTTP port, like `/metrics` — metadata only, scraped in-cluster by the controller.
- No proto change → no `grpc-contract.md` change. New CRD is additive/v1alpha1-safe.

## Test plan
- [x] `make pre-pr` green (naming, buf lint, fmt, vet, **race**, build, no generated drift)
- [x] `make cover-check` — logic coverage **79.6%** (≥65 gate)
- [x] Unit tests: index `Snapshot()` (dedup/sort/totals); server `/snapshot` end-to-end (ingest→JSON); controller `buildCacheIndexStatus`, `statusEqual` (timestamp-insensitive), `fetchSnapshot` (+non-200), `refresh` (fake client: create → no-op-on-unchanged → update-on-change)
- [ ] CI green on the PR
- [ ] (manual, later) `kubectl get cacheindex` on a kind cluster shows live aggregate


EdHasNoLife:
Addressed in fa77fe3.

- **Should-fix (create singleton before scrape):** fixed — `refresh()` now ensures the `cluster-default` CacheIndex exists *before* scraping `/snapshot`, so `kubectl get cacheindex` shows the CR (empty status) even when the server is unreachable; status fills on the next successful tick. New test `TestRefreshCreatesSingletonEvenWhenServerDown` covers the server-down path.
  - On the related point — the **default install doesn't deploy the server** (`config/default` only ships CRD/RBAC/manager): that's a real but separate gap (there's no server Deployment/Service manifest in `config/` yet at all). It's deployment scope, not this status-surface PR; the create-first fix makes the controller behave gracefully until the server is deployed. I'll file a follow-up ticket to add the server Deployment + Service to the install.
- **Should-fix (design doc):** updated `docs/design/grpc-contract.md` — the CacheIndex status surface + `/snapshot` are now documented as implemented (no longer a future follow-up).
- **Nit (server flag help):** `cmd/server` `--http-bind-address` help now lists `/snapshot`.

EdHasNoLife:
Addressed in 87effb8. Tightened the CacheIndex RBAC: the main resource is now `create;get;list;watch` (dropped `update;patch;delete`, which the poller never uses on the resource itself — it only writes the `status` subresource). Kept `list;watch` because the poller reads through the manager's **cached client**, where a `Get` is informer-backed and needs them; noted that on the marker. `cacheindices/status` stays `get;update;patch`.

EdHasNoLife:
Addressed in 76571cb.

- **Blocking (prefix shape):** fixed. Introduced `PrefixStatus{ Summary PrefixSummary }` so the status surface is `status.prefixes.summary.{total,hot}` — matching the locked contract — instead of the flattened `status.prefixes.{total,hot}`. The CRD now generates the nested shape, the print column points at `.status.prefixes.summary.total`, and the controller + tests were updated. Getting this right now avoids a breaking v1alpha1 field change later.
- **Should-fix (PROJECT metadata):** registered the cluster-scoped `CacheIndex` resource (`namespaced: false`) in the kubebuilder `PROJECT` file alongside `CacheBackend`.

EdHasNoLife:
Addressed in c626ecf.

- **Nit (operator RBAC):** added `cacheindex-viewer-role` and `cacheindex-editor-role` ClusterRoles (mirroring the CacheBackend convenience roles) and wired them into `config/rbac/kustomization.yaml`, so operators get RBAC for the new `kubectl get cacheindex` surface.
- **Nit (comment):** corrected the `CacheIndexPoller` doc — the server *exposes* the aggregate via `/snapshot` and the controller *scrapes* it (pull), rather than "server pushes."

EdHasNoLife:
Addressed in 0fc84e7.

- **Should-fix (hidden zeros):** dropped `omitempty` on `prefixes.summary.{total,hot}` so an empty index renders `total: 0` / `hot: 0` explicitly, matching the contract shape. New test asserts the zero summary serializes. (Left `omitempty` on the per-replica/tenant numeric fields — those are per-list-item optional metrics, not the always-present summary; happy to drop it there too if you'd prefer fully explicit zeros.)
- **Should-fix (lastUpdated semantics):** kept the write-on-change design (the locked decision) but made the field honest — `lastUpdated` is now documented as "when the aggregate last *changed*" (not last poll), and the print column is renamed **Updated → Changed** so operators don't misread it as poller freshness. Poller liveness is observable via the controller's own health/metrics; if you'd like an explicit freshness heartbeat on the CR, I can add one as a follow-up (it would mean a periodic write even when unchanged).

EdHasNoLife:
Addressed in e68b6ba — `fetchSnapshot` now wraps the `http.NewRequestWithContext` error with context (incl. the URL), consistent with the other wrapped returns in the function.

heymrbox (approved):

PR #17, opened 2026-05-27
Add full-chain CPU e2e canary for the cache substrate

EdHasNoLife:
## Summary

A one-shot, **GPU-free** canary that exercises the whole substrate end to end and asserts it works:

```
CPU vLLM engine ──ZMQ KV events──▶ kvevent-subscriber ──gRPC──▶ policy server ──▶ index
```

`docs/reference-stack/scripts/canary_e2e.sh` builds `server` + `kvevent-subscriber`, starts a CPU vLLM engine (with KV events), the policy server, and the subscriber, fires a repeated long prefix, then asserts:
1. an engine **prefix-cache hit** (`vllm:prefix_cache_hits_total` increases), and
2. the **index populated** end-to-end (`inferencecache_index_entries{model} > 0`).

It manages/cleans up the engine container, exits non-zero on failure, and is arch-aware (arm64/x86_64 image tag). Documented in the reference-stack README.

## How it's run

On-demand (not a blocking CI gate): it needs Docker, pulls the vLLM CPU image (multi-GB), and a Docker VM with ~10+ GiB RAM. Run locally, or wire into a scheduled/dispatch job on a Docker-capable runner.

## Verification

Ran locally end-to-end: engine healthy, requests 200/200, `prefix_cache_hits 0→2560`, `index_entries{model=canary}=20`, **PASS**.

## Note

Stacked on #15 (C1 — the `kvevent-subscriber` the full chain needs). Base is the C1 branch; will retarget to `main` once C1 merges. `make pre-pr` green.


EdHasNoLife:
Addressed the blocking find in 25b3a33: the canary now runs `make build` instead of `go build -o bin/server …`, so the gitignored `bin/` is created consistently with the rest of the repo and a fresh CI checkout (scheduled workflow) won't trip. Also rewrote the cleanup pid-kill into explicit `if`/`then` to clear the ShellCheck `A && B || true` nit. `bash -n` + `make pre-pr` green.

heymrbox (approved):

PR #24, opened 2026-05-28
Emit fail-open / degraded Events from CacheBackend (C4)

EdHasNoLife:
## Summary
- Wires `record.EventRecorder` into the CacheBackend reconciler and emits Kubernetes Events on transitions of the observed backend health and the effective `spec.integration.failOpen` toggle. Events fire **only on transitions**, never on steady state.
- Adds `spec.integration.failOpen *bool` (default `true` via `+kubebuilder:default=true`) + `status.failOpen` mirror so operators can read the current mode from status alone, and so transition detection has a stable previous-value baseline (missing echo ⇒ API default `true`, so first-apply `failOpen=false` correctly fires the Warning).
- Exports `IntegrationFailOpen(spec)` helper for the engine adapter (spec → engine connector flag in C6).

The phase enum + `Ready` condition derivation already live in C2's `managedHealth`; this change only layers events + `failOpen` plumbing on top.

## Event reasons
| Reason | Type | When |
|---|---|---|
| `BackendDegraded` | Warning | health enters Degraded (message names the failure mode, e.g. `0/2 replicas available`) |
| `BackendRecovered` | Normal | health leaves Degraded for Ready |
| `FailClosedEnabled` | Warning | `failOpen` flipped `true → false` — cache becomes a serving dependency (advanced opt-in) |
| `FailOpenRestored` | Normal | `failOpen` flipped `false → true` |

Linear: [CAC-24](https://linear.app/cachebox/issue/CAC-24/c4-m5-fail-open-semantics-degraded-status-events).

## Composition with neighbouring tickets
- **C3 (CAC-23)** — touches the same `Reconcile` + `patchStatus` while it's in flight; the conflict is structural (both add helpers around the same dispatch), not semantic. C3's `Progressing` condition + HPA composes cleanly with this PR's Recorder + transition-emit.
- **C6 (CAC-26)** — owns `InjectEngineConfig`. This PR ensures `failOpen` exists, defaults to true, and is surfaced; C6 will consume `IntegrationFailOpen(spec)` to set the connector flag on engine pods.
- **C7 (CAC-27)** — admission validation lives in a separate webhook; not touched here.
- Per-request fail-open serving behavior at the engine level remains a vLLM+LMCache property configured via flags (out of scope).

## Test plan
- [x] `make ci` green (verify-naming + fmt + vet + race tests + build).
- [x] `make pre-pr` green (no generated-code drift).
- [x] Coverage: 81.3% (gate 79%).
- [x] Unit tests in `internal/controller/cachebackend_events_test.go`:
  - Ready → Degraded → Warning `BackendDegraded` recorded; message names the failure mode.
  - Degraded → Ready → Normal `BackendRecovered` recorded; no spurious second Warning.
  - 5 steady-state Ready reconciles → 0 events emitted.
  - First-apply with `failOpen: false` → Warning `FailClosedEnabled` ("serving dependency"); status echoes `false`; second reconcile is silent (idempotent).
  - Flip back to `failOpen: true` → Normal `FailOpenRestored`; status echoes `true`.
  - Default `failOpen` (nil) → status echoes `true`; no events.
  - Nil `Recorder` is safe (defence-in-depth).
- [x] `TestIntegrationFailOpenDefaultsTrue` exercises the helper.
- [x] CRD-schema test asserts `integration.failOpen` is `boolean` with `default: true` and `status.failOpen` is present.

Envtest coverage is intentionally not added in this PR — fake-recorder coverage is high-fidelity for the behavior under test (transition detection, reason codes, message content). Can follow up if reviewers want apiserver-mediated defaulting coverage.


EdHasNoLife:
Re: latest Codex review — appreciated, but I'm intentionally leaving the contract as designed in this PR. CAC-24's Definition of Done explicitly requires "**Fail-closed only via explicit opt-in.**" so providing the opt-in is the ticket's scope, not a contract drift.

The core invariant ("the cache is an optimization, never a serving dependency") is preserved:
- `spec.integration.failOpen` defaults to `true` via `+kubebuilder:default=true`, so any client that omits the field gets the unconditional fail-open contract Codex is protecting.
- Setting `failOpen: false` is loud and visible: the controller emits a Warning `FailClosedEnabled` Event on every transition, and `status.failOpen` echoes the observed value so the cluster operator can never silently flip a backend into "serving dependency" mode.

`docs/design/cachebackend-api.md` is being updated in the same PR so the design doc reflects what the CRD now exposes (per project rule that API changes ship with their doc). The previous wording ("fail open unconditionally") predated this ticket's planned opt-in.

A per-engine `backendConfig` knob would have been an alternative, but the spec deliberately puts this under `spec.integration` so the controller can surface it in status + events (which a free-form opaque backendConfig string couldn't). C6 will consume `IntegrationFailOpen(spec)` to plumb the bool into the engine connector flags.

Treating this as informational; not blocking.

heymrbox (approved):

PR #29, opened 2026-05-28
Implement CachePolicy, CacheTenant, PromptTemplate, PDTopology, CacheIndex

heymrbox:
## Summary

Adds the remaining cache policy-surface CRDs for CachePolicy, CacheTenant, PromptTemplate, PDTopology, and status-only CacheIndex. This includes API types, generated deepcopy/CRD manifests, RBAC roles, sample manifests, schema/deepcopy tests, and design docs for the new CRD contracts.

## Linked issues

Closes CAC-16

### Quality
- [ ] CI is green.


api/v1alpha1/cachepolicy_types.go
```diff
@@ -0,0 +1,121 @@
...
+type CachePolicyTenantScopeType string
+
+const (
+	CachePolicyTenantScopeAll       CachePolicyTenantScopeType = "All"
+	CachePolicyTenantScopeTenant    CachePolicyTenantScopeType = "Tenant"
+	CachePolicyTenantScopeNamespace CachePolicyTenantScopeType = "Namespace"
+)
+
+// CachePolicySpec defines cache lookup and eviction policy.
+type CachePolicySpec struct {
```
EdHasNoLife, line 19: **Should-Fix — CAC-55's DoD requires the index TTL + lookup-budget wiring, not just the CRD; this PR delivers the type but no consumer in pkg/index or pkg/server.**

Applying a CachePolicy with `ttl: 5m`, `minimumPrefixTokens: 512`, `lookupTimeoutMs: 20` has no effect: the index keeps using `DefaultTTL = 30m`, and `LookupRoute` ignores both `minimumPrefixTokens` and `lookupTimeoutMs`. Operators get a configuration surface with no behavior change. Recommend splitting CAC-55 into 'CRD types' (closed by this PR) and 'wiring + propagation' (follow-up ticket) so the gap is tracked rather than silently dropped.

api/v1alpha1/cachepolicy_types.go
```diff
@@ -0,0 +1,121 @@
...
+	// Eviction is the cache eviction algorithm.
+	// +optional
+	Eviction CachePolicyEvictionAlgorithm `json:"eviction,omitempty"`
+
+	// TTL is the maximum time a cache entry should remain usable.
+	// +optional
+	TTL *metav1.Duration `json:"ttl,omitempty"`
```
EdHasNoLife: **Should-Fix — field name drift from tech spec § 4.1 and the CAC-55 ticket: spec/ticket use `evictionTTL`; this field is `ttl`.**

A user copying the CachePolicy example from the tech spec (`evictionTTL: 30m`) finds the field unknown and is rejected by CRD validation. Either rename to `evictionTTL` to match the spec, or update the tech spec to reflect the split (`eviction` algorithm enum + `ttl` duration) and add an alias note in the design doc.

api/v1alpha1/cachepolicy_types.go
```diff
@@ -0,0 +1,121 @@
...
+	LookupTimeoutMs *int32 `json:"lookupTimeoutMs,omitempty"`
+
+	// FailOpen keeps inference serving when the cache is unreachable.
+	// +optional
+	// +kubebuilder:default=true
+	FailOpen *bool `json:"failOpen,omitempty"`
+
+	// TenantScope restricts where this policy applies.
+	// +optional
+	TenantScope *CachePolicyTenantScopeSpec `json:"tenantScope,omitempty"`
```
EdHasNoLife: **Should-Fix — `TenantScope` is a struct ({type, tenantRef, tenantID, matchLabels}); tech spec § 4.1 specifies `tenantScoped: true` (a bool).**

Users following the tech spec write `tenantScoped: true` and get a CRD validation error. The richer struct is arguably better, but it's a deliberate deviation that should be ratified in the tech spec (and CAC-55) before merging — otherwise the contract drifts silently from its published shape.

EdHasNoLife (commented):
**Reviewed against CAC-55 and CAC-16.** Personally ran `go build`, `go vet`, `go test ./api/v1alpha1/... -race`, `make manifests generate` (no drift), and `make pre-pr` — all green on `cdf118d`. Vendor-neutral check clean across all new files. Verified the `CacheIndex` `Spec` value→pointer change is safe (no caller on main writes `.Spec`).

### Coverage vs the tickets

**CAC-16 (B2 — remaining CRDs): fully addressed.** All five CRDs land with sensible kubebuilder markers; `TestRemainingCRDSchemas` + `TestRemainingCRDDeepCopies` cover schema (enums/defaults/minimums/required/list-map keys) and deepcopy round-trips for every kind; CacheIndex is enforced status-only via the `!has(self.spec)` XValidation rule and the Go `Spec` becoming an `*omitempty` pointer (nice touch — closes a CAC-16 DoD item that CAC-50 left implicit).

**CAC-55 (CachePolicy CRD): only the CRD half.** The DoD has four bullets; this PR delivers one:
- [x] `CachePolicy` CRD + manifests committed; vendor-neutral.
- [ ] Index eviction TTL driven per-namespace by `evictionTTL` — no `pkg/index` change.
- [ ] `LookupRoute` honors `minimumPrefixTokens` (`NO_HINT`) and `lookupTimeoutMs` (`TIMEOUT`) — no `pkg/server` change.
- [ ] Propagation mechanism implemented + tested — no controller/server wiring.

Recommend: split CAC-55 — this PR closes the CRD half; a follow-up ticket tracks index/lookup wiring + the controller-server propagation. Update CAC-55 to reflect the split so the wiring isn't silently dropped.

### Inline findings
Three Should-Fix items below (the wiring gap + two naming drifts from the tech spec § 4.1 / CAC-55 ticket).

### Notes (not findings)
- PR title says "Implement … CacheIndex"; the CacheIndex change is a 3-line annotation + pointer tweak. Title could be sharper.
- `SchemeBuilder.Register(func(s *runtime.Scheme) error {...})` is more verbose than the existing CacheBackend pattern (`SchemeBuilder.Register(&Type{}, &TypeList{})`). Both work; consistency is a small win.

heymrbox:
Addressed in 57e5f9a.

I made the CacheIndex validation compatible with existing v1alpha1 objects by allowing either no spec or the legacy empty `spec: {}` shape, while keeping the API status-only.

I also registered CachePolicy, CacheTenant, PromptTemplate, and PDTopology in `PROJECT`, added `MinLength=1` validation for the required identifier/body fields, regenerated the CRDs, and updated the schema tests and design doc.

Verified with:
- `go test ./api/v1alpha1`
- `make pre-pr`
- pre-push `make ci` gate

heymrbox:
Addressed the latest Codex Review in 0e6fd8a.

I restored `CacheIndex.Spec` to the value-type Go API so v1alpha1 clients are not source-broken, while keeping the CRD validation compatible with both omitted spec and legacy empty `spec: {}`.

I also made root `spec` required for `CacheTenant` and `PromptTemplate`, regenerated the CRDs/deepcopy output, and added schema assertions for those root requirements.

Verified with:
- `go test ./api/v1alpha1`
- `make pre-pr`
- pre-push `make ci` gate

api/v1alpha1/cachepolicy_types.go
heymrbox: Aligned the public contract in `80b679d`.

`spec.ttl` is now `spec.evictionTTL` in the Go type, generated CRD, sample manifest, and design doc so the published example shape matches what the CRD accepts.

Verification:
- `go test ./api/v1alpha1 -run TestRemainingCRD(Schemas|DeepCopies)$`
- repo CI gate via pre-push hook: `make ci`

api/v1alpha1/cachepolicy_types.go
heymrbox: I kept the runtime-wiring gap explicit instead of pretending this PR closes it.

This update narrows the PR to the declarative CachePolicy contract and the design doc now says the server/controller propagation work is follow-up scope. I did not add pkg/index or pkg/server consumers in this patch because there is still no policy-loading seam in the server today.

Verification on the pushed branch (`80b679d`):
- `go test ./api/v1alpha1 -run TestRemainingCRD(Schemas|DeepCopies)$`
- repo CI gate via pre-push hook: `make ci`

api/v1alpha1/cachepolicy_types.go
heymrbox: Aligned this to the published contract in `80b679d`.

The public shape is now `tenantScoped: true` rather than the richer selector object. That change is reflected in the Go type, generated CRD, sample manifest, and schema/deepcopy coverage so users following the spec no longer hit validation drift.

Verification:
- `go test ./api/v1alpha1 -run TestRemainingCRD(Schemas|DeepCopies)$`
- repo CI gate via pre-push hook: `make ci`

EdHasNoLife (commented):
**Round 2 — re-reviewed `4bfd026` against round-1 findings.**

All three round-1 items are addressed (checked by inspection of the latest diff, the design doc update, and the Codex thread; I did not re-run tests locally because my worktree is on another branch — leaning on the existing CI + Codex passes):

- ✅ `spec.ttl` → `spec.evictionTTL` (matches tech spec § 4.1 + CAC-55).
- ✅ `tenantScope` struct → `tenantScoped` boolean (matches tech spec).
- ✅ Scope gap noted in `policy-crds.md`; CachePolicy runtime wiring tracked in CAC-62.

The `CacheIndex.Spec` value→pointer change Codex flagged as a v1alpha1 source-break — that catch was correct (round 1 missed the external-consumer angle), and the resolution (keep value-type Go API + `!has(self.spec)` CEL on the CRD) is clean.

For tracking hygiene around the CRDs that don't yet have consumers: CAC-16 is now In Progress with PR #29 attached, and the wiring follow-ups are split out — **CAC-62** (CachePolicy enforcement), **CAC-66** (CacheTenant quota + status writer), **CAC-69** (PDTopology reconciler + LookupPDRoute Phase 2). So nothing is silently dropped on merge.

**Verdict: Approve.** No new inline findings.

heymrbox:
Addressed in e5b3619.

Changes:
- Restored CacheIndex v1alpha1 compatibility by removing the strict empty-spec admission schema. Legacy non-empty `spec` fields are pruned and accepted instead of becoming a new rejection path.
- Kept `CacheTenant.spec.crypto` as a reserved field, but made non-empty config reject explicitly with `maxProperties: 0` plus preserved unknown fields.
- Added generated-CRD schema/admission tests for both behaviors and regenerated the CRDs.

Verified:
- `go test ./api/v1alpha1 -run 'TestRemainingCRDSchemas|TestGeneratedCRDAdmissionValidation'`
- `go test ./api/v1alpha1`
- `make pre-pr`
- pre-push `make ci`

EdHasNoLife (approved):

PR #37, opened 2026-05-29
Register gRPC server reflection on the policy server (CAC-49)

EdHasNoLife:
## Summary

The policy server registered `InferenceCache` + `grpc.health.v1` but not gRPC server reflection, so `grpcurl list` / `describe` / `grpc_health_probe` required hand-passing `-import-path proto -proto …/inferencecache.proto`. For a deployable server that operators and engine/gateway integrators will poke at, that's a poor first impression.

This PR wires `reflection.Register(grpcServer)` next to the existing service registrations in `pkg/server/server.go`, so the wire is now self-describing. Reflection exposes only the service schema — never KV or prompt data.

## Changes
- `pkg/server/server.go`: import `google.golang.org/grpc/reflection`; call `reflection.Register(grpcServer)` in `New()` alongside the InferenceCache + health registrations.
- `pkg/server/server_test.go`: new `TestGRPCServerReflectionEnumeratesRegisteredServices` drives a `ServerReflectionInfo` stream against the bufconn-backed server and asserts `ListServices` enumerates both `inferencecache.v1alpha1.InferenceCache` and `grpc.health.v1.Health` — a regression that fails to register either service is caught by `make ci`. The existing `startInProcessServer` is split into a thin typed-client wrapper around a new `startInProcessServerConn` so this test (and any future non-InferenceCache test) can attach its own client to the same wire.

## Out of scope (deliberately deferred)
- **Reflection on/off flag.** The ticket mentions a `Config` field as a *consideration* for hardened prod. Left out to avoid a config-surface decision we don't need now — reflection exposes only the schema, no data. Trivial to add later if a hardened-prod profile lands.
- **TLS** on the gRPC server (post-Phase-1).
- **Auth/authz** on the reflection RPC (Phase 2).
- **`grpc_health_probe` readiness probes** in the controller's deployment manifests for the server (tracked separately).

## Test plan
- [x] `go test -race ./pkg/server/...` — passes, including the new wire test.
- [x] `make ci` — race tests + lint + build green.
- [x] `make pre-pr` — generated-code drift clean.
- [ ] Manual smoke against a running server (would require a built binary + a port; the wire test is the gate):
  - `grpcurl -plaintext localhost:9090 list` enumerates both services with no `-proto` flag.
  - `grpcurl -plaintext localhost:9090 describe inferencecache.v1alpha1.InferenceCache` returns the service descriptor over the wire.
  - `grpcurl -plaintext localhost:9090 grpc.health.v1.Health/Check` returns `SERVING`.


EdHasNoLife:
**Scope expansion (CAC-49 description updated).**

CI/CD audit found several gRPC endpoints lack wire-level tests — handler-only or none. The `startInProcessServerConn` bufconn helper this PR already introduces is the natural home for the missing coverage, so we're folding it into CAC-49 instead of opening a separate PR. Please add wire tests for the following before merge:

- `GetCacheState` — over-the-wire aggregate assertion (replicas + summary).
- `PublishEvent` — over-the-wire `CacheEvent` ingest → assert `Ack` + index mutation; at least one cross-scheme negative case.
- `StreamCacheEvents` — open stream, assert clean EOF (pins the stub close-immediately contract so a future real implementation can't silently regress to error/hang).
- `StreamMetrics` — same as above, pins the stub contract.
- `grpc.health.v1.Health/Check` — over-the-wire `SERVING` assertion (this is what `grpc_health_probe` will use as the K8s readiness gate once the server's default-install Deployment lands).

Each test is ~10–20 lines, reuses the bufconn helper already in the PR, and doesn't touch any production code beyond what's already in the diff. Existing wire coverage for `LookupRoute` + `ReportCacheState` must stay green. Full rationale and updated DoD on the ticket.

heymrbox (approved):

PR #43, opened 2026-05-29
DO NOT MERGE: deliberate-failure repro for install-smoke gate

EdHasNoLife:
**Do not merge.** This branch deliberately drops `- namespace.yaml` from `config/default/kustomization.yaml` so the `default-install-smoke` workflow has the exact regression class the gate is meant to catch (the missing-Namespace install regression that previously slipped past the heavier canaries).

Linked PR for the actual gate: #42.

Expected outcome:
- `default-install-smoke` fails on `kubectl apply -k config/default` with a `namespaces "inference-cache-system" not found` error from namespace-scoped object creates.
- The on-failure `actions/upload-artifact` step uploads `/tmp/install-smoke-logs/` containing pod state + `kubectl describe` output, proving the diagnostics path works on an early-exit failure (not just `fail()`-routed paths).

Will be closed once the artifact is captured and pasted into #42's description.


EdHasNoLife:
Closing — install-smoke caught the deliberate regression as expected (red in 3m17s, failure-artifact pasted into #42). Branch deleted; not for merging.

PR #4, opened 2026-05-27
Add Codex code-review workflow for pull requests

EdHasNoLife:
## Summary
- Add `.github/workflows/codex-review.yml` — an advisory PR code review using the official `openai/codex-action@v1`.
- The Codex prompt encodes this project's review standard: vendor-neutral naming, design-doc/proto↔doc sync, fail-open gRPC semantics, v1alpha1 backward-compatibility, and test coverage.
- Result is posted back as a PR comment.

## Design
- **Advisory, separate from `ci.yml`** — an LLM review never hard-blocks the build/test merge gate.
- Runs `sandbox: read-only`; uses `pull_request` (not `pull_request_target`) and skips forked PRs, which lack secret access.

## Setup required
- Add an `OPENAI_API_KEY` repository secret (Settings → Secrets and variables → Actions). Without it, the `codex` job fails on same-repo PRs.

## Test plan
- [x] Add `OPENAI_API_KEY` secret
- [ ] Confirm the workflow runs on this PR and posts a review comment
- [ ] Confirm forked PRs skip the job cleanly (no red X)

PR #48, opened 2026-05-30
codex-review workflow: tune prompt for exhaustive per-round findings

EdHasNoLife:
## Summary

Replace the `prompt:` block in `.github/workflows/codex-review.yml` to encourage comprehensive per-round findings instead of round-to-round drift. Single contiguous edit; no other workflow surface touched (concurrency, fork-skip guard, action SHAs, `post_feedback` job all untouched).

## Diagnosis of the round-to-round drift

The current prompt drifts for six specific reasons:

1. `Be concise and specific` implicitly tells the model to trim output — concise + comprehensive are in tension, so findings get dropped.
2. No "walk every category, every time" instruction — the four numbered checks are topics to think about, not a checklist to confirm. The model can satisfy the prompt with two strong findings and stop.
3. No "treat as fresh review" framing — on re-review, the model has no instruction telling it that fix pushes are not a license to roam to new territory.
4. No file-enumeration step — "Read the full diff" lets the model pick its own scope; some files never get opened until next round.
5. `If there are no issues, say so briefly` is a "skip the category if clean" license — there's no per-category coverage report.
6. No exclusion of generated code — wasted finding slots on `zz_generated*` / `pkg/server/proto/` displacing real findings.

## New prompt's behavioral changes

- **File-enumeration step** (STEP 1) — every changed file is listed up front and either reviewed or marked `[skipped — generated]`, so scope is explicit.
- **Per-category coverage report** — each of the four checks is reported as `Clean` or with file:line references, so a clean category is positively asserted, not silently omitted.
- **"Treat as fresh, cold review" framing** — counters drift on re-review after fix pushes.
- **Exhaustive enumeration explicitly invited** — the rationale per finding stays terse, but the finding list is no longer trimmed for brevity; nits are kept rather than dropped.
- **Generated-code exclusion** — `config/crd/`, `config/rbac/role.yaml`, `pkg/server/proto/`, `zz_generated*` are explicitly skipped so finding slots aren't wasted on regenerated noise.

## Test plan

- [x] YAML parses (`ruby -ryaml -e 'YAML.load_file(...)'`).
- [x] All six GitHub Actions expressions preserved (`pull_request.number`, `repository`, `base.sha`, `head.sha`, `title`, `body`).
- [x] Workflow structure intact: concurrency, fork-skip guard (`head.repo.full_name == repository`), action SHAs (`checkout@de0fac2`, `codex-action@e0fdf01`, `github-script@3a2844b`), `post_feedback` job all untouched.
- [x] `make pre-pr` passes (verify-naming + verify-no-internal-refs + fmt + vet + lint + race tests + build + drift gate + coverage gate ≥ 79%).
- [ ] **Observational: watch the next 2-3 real PR reviews.** Goal — round 1 surfaces a comprehensive set of findings; round 2 (after fixes) finds little or nothing new. If round-to-round drift persists, file a follow-up to investigate the non-prompt levers (temperature, model pinning).


EdHasNoLife:
Re: Codex Should-fix on `.github/workflows/codex-review.yml:78` (generated-files skip).

Declining for this PR — the prompt structure is locked by the ticket and this change is deliberately scoped to a pure rewrite. Two reasons the tension is acceptable in the current shape:

1. **Quality check (#4) already covers the failure mode.** The checks block explicitly says "generated code (config/crd, zz_generated*, pkg/server/proto) is regenerated, not hand-edited" — so a PR that touches generated files without their source counterparts should still surface as a Quality finding even though STEP 1 doesn't open the file line-by-line.
2. **CI has independent guards.** `make pre-pr` runs the generated-code drift check (`git diff --exit-code -- config api`) and proto-gen drift in CI — a hand-edit to generated files that doesn't match the sources fails the build before review runs.

If we see this fail in practice (Codex misses a hand-edit, or the Quality check doesn't trigger), I'll file a follow-up to tighten STEP 1 — but I'd rather observe the new prompt's behavior over 2-3 PRs first per the ticket's test plan.

heymrbox (approved):

PR #51, opened 2026-05-30
Secure /snapshot with bearer-token auth + NetworkPolicy

EdHasNoLife:
## Summary

Splits the policy server's internal `/snapshot` endpoint onto its own listener (default `:8081`), gates it with TokenReview-backed bearer auth, and ships a namespace-scoped NetworkPolicy. The CacheIndex poller sends its projected ServiceAccount token on every scrape. `/healthz`, `/readyz`, `/metrics`, `/policy` stay open on `:8080` — kubelet probes and Prometheus scrapes unchanged.

## What changed

- **Server** — new `--snapshot-bind-address` flag + dedicated `http.Server` for `/snapshot`. New `pkg/server/auth` middleware validates `Authorization: Bearer` via the apiserver's TokenReview API, admits only `--snapshot-allowed-sa`, caches successful results 30s by SHA-256 of the token, fails-closed on apiserver errors (503), and reports outcomes as `inferencecache_snapshot_auth_total{result=ok|unauth|forbidden|error}`.
- **Default install** — snapshot port added to the Service; server SA bound to `system:auth-delegator` so it can create TokenReviews; projected SA token automounted on the server pod.
- **NetworkPolicy** — namespace-scoped, pod-selector based; allows ingress to `:8081` only from pods matching the controller selector. `:8080` is NOT covered (kubelet + Prometheus need it).
- **Controller poller** — re-reads `/var/run/secrets/kubernetes.io/serviceaccount/token` on every scrape (tmpfs, handles kubelet rotation), sends it as a bearer. Existing fail-soft path covers 401/403/network errors.
- **Tests** — middleware unit table-driven (missing/empty/invalid/wrong-SA/apiserver-error/cached-token + LRU eviction); envtest end-to-end (real apiserver mints TokenRequests, controller SA → 200, other SA → 403, no token → 401); split-listener regression check (`/snapshot` must NOT serve on `:8080`); `default-install-smoke` extended with a side pod that GETs `/snapshot` without auth and asserts 401 / NetworkPolicy drop.

## Test plan

- [x] `make ci` (verify-naming, verify-no-internal-refs, fmt, vet, lint, race tests, build) — green
- [x] `make pre-pr` (ci + generated-code drift) — green
- [x] `make cover-check` — logic coverage 87.7% ≥ 79%
- [ ] `default-install-smoke` on the PR runner — exercises NetworkPolicy + 401 from a side pod against a real kind cluster
- [ ] envtest suite runs locally with `KUBEBUILDER_ASSETS` set

## Threat model

Bearer over in-cluster cleartext mirrors the shape kubelet itself uses to ship SA tokens — acceptable for the in-cluster trust boundary. mTLS for `/snapshot` is a possible future hardening and is out of scope here (a separate gRPC-TLS posture decision lives in its own ticket).

## Backward compat

`/snapshot` is internal; the only consumer is the CacheIndex poller, which is updated in this same change. No external clients.

## Coordination

- Hardens the endpoint shipped by the CacheIndex poller introduced earlier — that poller is the consumer being updated here.
- Ships alongside the default-install manifests; NetworkPolicy lives under `config/server/` and rides into `config/default` via the existing kustomization.
- A separate `indexParticipation` writer in flight reads from the same `/snapshot` path — once both land it inherits this auth transparently.
- The `default-install-smoke` kind gate is the e2e home for the new NetworkPolicy / 401 assertions.

## Stacking note

Currently stacked on `cac-90-bump-golangci-lint` (open PR, all checks green, awaiting approval) — that branch carries the `golangci-lint v2.12.2` bump this PR's `make ci-lint` depends on. Once that lands, I'll rebase onto `main` and the lint commits will dedupe from the diff.


EdHasNoLife:
Round-1 Codex findings — responses + push notes:

**Blocking (NetworkPolicy `:8080`/`:9090` drop)** — false positive. The manifest has two ingress rules: rule 1 opens `8080` + `9090` with no `from:` filter (preserves the pre-policy "any in-cluster client" posture for kubelet probes, Prometheus scrapes, `/policy` push, gRPC API), rule 2 restricts `:8081` to the controller selector. The header comment on `config/server/networkpolicy.yaml` (lines 11–19) explicitly calls out this CNI-isolation footgun and how the dual-rule shape addresses it. Codex appears to have stopped at the second rule.

**Should-fix (`docs/design/policy-propagation.md` drift)** — false positive. The doc was updated in the same commit (`60b14fd`) — see lines 21–29: `/policy sits on the server's open HTTP port (:8080)… /snapshot is on a separate internal listener (:8081) gated by TokenReview-backed bearer auth and a NetworkPolicy…`. No drift.

**New in `0645c4e`** — addressing reviewer feedback on test depth: added `internal/controller/cacheindex_authed_integration_test.go`, an envtest that drives the full **client-to-server-to-apiserver** path:

```
CacheIndexPoller.refresh
  -> bearerToken() reads tmpfile (kubelet projection shape)
  -> fetchSnapshot() sets Authorization: Bearer <token>
  -> pkg/server/auth.Middleware
  -> TokenReview against the envtest apiserver
  -> handler returns synthetic index.Snapshot
  -> poller writes CacheIndex.status
```

Cases covered: authorized SA → CR populated; wrong-SA → 403 + fail-soft status preservation; empty token → 401 + fail-soft preservation; recovery to a valid token converges. The middleware-level envtest (`pkg/server/auth/integration_test.go`) covers the auth layer in isolation; this new test pins the production client glue on top. Both inherit CI envtest coverage via the existing `Install envtest binaries` step in `ci.yml`. The kind-based smoke (`default_install_smoke.sh`) already covers the positive auth path implicitly (via `cacheindex.status.observedServer` populating, which only succeeds if the authed scrape works through the NetworkPolicy + middleware) and the negative path explicitly (401 / NP-drop from a side pod).

heymrbox (approved):

EdHasNoLife:
Round-7 Codex findings — responses + push notes:

**Nits — fixed in 7419bfc.**

- `pkg/server/auth/middleware.go:215` — `extractBearer` now compares the auth scheme with `strings.EqualFold` so `Bearer` / `bearer` / `BEARER` all parse. Matches RFC 7235 §2.1. Added two table cases (lowercase + uppercase) in `middleware_test.go`.
- `docs/reference-stack/scripts/default_install_smoke.sh:267` — added a `kubectl delete pod --ignore-not-found --wait=true` before `kubectl run`, and now check the create's exit code instead of swallowing it with `|| true`. A leftover `ic-snapshot-probe` from an interrupted run no longer causes the script to silently read stale logs.

**Should-fix — both intentionally deferred to follow-up tickets; not landing in this PR.**

- `internal/controller/cacheindex_controller.go:40` / `pkg/server/auth/middleware.go:161` (audience binding) — tracked as a deliberate scope split in a follow-up ticket (parent: this PR's ticket). The default-audience SA token is the same shape kubelet uses for its own probes; tightening to a snapshot-bound audience requires a coordinated change across (a) the controller's projected-token volume in the manager Deployment, (b) the `TokenReviewSpec.Audiences` check in the middleware, and (c) RBAC for the new audience name. That's its own PR — both the projection and the validation must flip in lockstep or the controller starts getting 403s in production.

- `docs/reference-stack/scripts/default_install_smoke.sh:292` (smoke accepts 401 as success → NetworkPolicy enforcement untested) — real and known. kind's default kindnet CNI does not enforce NetworkPolicy, so the L3/L4 gate cannot be exercised by the current smoke runner. Validating it in CI requires swapping kindnet for an enforcing CNI (Calico / Cilium / Antrea), tightening the assertion to require `curl_failed:*` (not `401`), and adding a positive-path probe (controller-labeled pod without a token → reaches `:8081`, gets `401`). That's a foundational change to the smoke runner — CNI choice affects every future PR's runtime budget — so it's a separate ticket rather than buried in this one. The bearer-auth gate is the primary security boundary and IS fully exercised here (unit table + envtest end-to-end against a real apiserver + smoke negative probe).

Both follow-ups are filed in Linear; the PR description's coordination section can be updated to call them out if reviewers want.

EdHasNoLife:
Round-8 — all findings addressed inline in d8738aa.

- **Should-fix** (smoke `curl_failed:*` too broad) — case now requires `curl_failed:28` (curl's timeout exit code, which is the shape of an enforcing-CNI silent drop). Other failure modes — DNS (6), ECONNREFUSED (7), malformed URL (3) — now fail the smoke instead of masking regressions. Header comment carries the full exit-code rationale. CAC-112 (already filed) covers the next step: swap kindnet for an enforcing CNI and drop the `401` branch entirely, so the smoke proves the NetworkPolicy is doing the work rather than the auth.
- **Nit** (`/snapshot` accepts non-GET) — handler now returns 405 with `Allow: GET` for POST/PUT/DELETE/PATCH. New test `TestSnapshotRejectsNonGet` pins the contract.
- **Nit** (`middleware.go:187` says "controller's log") — reworded to "server log" with an inline note that the middleware runs in the policy-server process. Operators chasing the diagnostic now look in the right pod.

EdHasNoLife:
Round-9 — all 4 findings addressed inline in ecef6b1.

**The 3 Should-fix items are the same underlying issue: `TokenReview.Status.Error` was incorrectly collapsed onto the routine-401 path.** Codex is right and my earlier rationale was empirically wrong — kube-apiserver leaves `Status.Error` *empty* for plain bad tokens; it populates the field when the authenticator chain itself fails to answer (webhook timeout, parse error, downstream cert failure). Those two shapes deserve different metric labels because they drive different operator responses.

- **middleware.go**: in the `!Authenticated` branch, split on `Status.Error`. Empty → `result="unauth"` / 401 (routine bad token). Populated → `result="error"` / 503 (authenticator-chain fault), with the `Status.Error` text logged at WARN. Both still deny; the difference is the alert surface.
- **middleware_test.go**: flipped the existing `status-error` table case from `ResultUnauth/401` to `ResultError/503`. Rewrote its docstring to explain the corrected mental model. Updated the fake reviewer's `Status.Error` string to "webhook authenticator timeout" so the fixture matches the documented shape.
- **metrics.md**: rewrote the `inferencecache_snapshot_auth_total{result}` semantics. Now spells out the unauth-vs-error split with operator guidance — alert on `unauth` → controller pod; alert on `error` → apiserver/authenticator chain. This was the missing operational contract.

**Nit** (cacheindex_controller.go:422): escaped quotes (`\"\"`) in the bearerToken godoc were copy-paste artifacts. Fixed to literal `""`. Pure comment polish.

No deferrals this round.

EdHasNoLife:
Round-10 — round-9's `Status.Error` split was empirically wrong; reverted in 2071306.

The envtest integration `TestAuthMiddleware_AgainstEnvtestAPIServer/bogus_bearer_->_401` (which runs the middleware against a real kube-apiserver, not a fake reviewer) caught it: a literal `"not-a-real-token"` returns `Authenticated=false` with `Status.Error` *populated* — the SA-token authenticator's JWT parser produces an error string for any input it cannot decode.

So Codex's underlying claim ("`Status.Error` is for authenticator-chain faults") is half-true: kube-apiserver uses the field for **both** (a) authenticator-chain faults AND (b) routine bad-token cases. The response shape is the same; there is no machine-readable discriminator. Parsing the error string to split them would be brittle and version-specific.

Reverted to the round-8 behavior — `!Authenticated` → 401 / `result="unauth"` regardless of `Status.Error` — but with the comment expanded to record the round-9 finding so the next reviewer doesn't try the same well-meaning split. The middleware still logs `Status.Error` text at WARN so the operator can disambiguate via the server log; the metric is intentionally lossy.

The only path that maps to `result="error"` / 503 is now the Go-level `CreateTokenReview` error (transport failure, RBAC reject before review even ran). That distinction IS robust because it triggers on a wire-level fault before the apiserver responds — it cannot collide with any TokenReview status shape.

`metrics.md` updated to match — `result="unauth"` semantics now document the collapsed shape and the operator-response model (read the WARN log to disambiguate). Nothing else from rounds 7-9 reverts.

This is exactly why the envtest gate exists. The unit-level fake reviewer didn't catch the disagreement because it was free to define its own behavior; the real apiserver disagreed.

EdHasNoLife:
Round-11 Should-fix — same audience-binding finding as round-4 (re-surfaced after the listener split + middleware churn). Already triaged: filed as a self-contained follow-up rather than expanding this PR, because audience binding is a 3-component change (controller Deployment volume projection + `--snapshot-audience` flag on the server + `TokenReviewSpec.Audiences` in the middleware) plus two new test cases (envtest wrong-audience → 401, smoke default-audience side-pod → 401). Diff would grow ~25%; the existing gate (bearer auth + NetworkPolicy + TokenReview SA validation) works as specified by the original threat model.

The defense-in-depth value is real but layered on top of, not part of, CAC-68's specified gate. Deferring is the right altitude call.

PR #53, opened 2026-05-31
Wire CacheTenant quota enforcement + status writer into the index

EdHasNoLife:
## Summary
Wire CacheTenant quota enforcement + status writer into the index. Enforce per-tenant index-entry budgets at ingest and surface live usage on `CacheTenant.status`.

Memory budgets are scoped out: the engine KV cache is a shared, tenant-unaware pool, so the control plane can neither enforce nor honestly attribute bytes per tenant (`cache_memory_bytes` is the engine total, double-counted across tenants on a shared engine). `maxIndexEntries` — the one cache resource the index owns — is enforced.

## What changed
- **CRD**: removed `spec.quota.maxMemoryBytes` + `status.memoryUsed`; kept `maxIndexEntries`. Printer columns Tenant/Entries/Quota/Isolation/Age.
- **/policy snapshot**: extended with `[]ResolvedTenant`; `PolicyPropagationVersion` 1 → 2.
- **Index**: ingest-time `evictOldestForTenant` (Fairness — a tenant evicts only its own oldest entries down to budget) via a `TenantQuotaResolver`; fail-open when no quota. Metric `inferencecache_tenant_evictions_total{tenant_id,reason}`. Snapshot now carries per-tenant `indexEntries`.
- **Aggregate invariant**: per-tenant `indexEntries` and the cluster total are computed from a single `Aggregate()` walk under one lock, so `Σ tenants[].indexEntries == prefixes.summary.total` holds by construction (untenanted entries bucket under a reserved `_default` sentinel). Table-driven test across empty / single / multi-tenant (incl. multi-replica prefix) / default-tenant / post-eviction.
- **Controller**: one reconciler (renamed `ControlPlaneReconciler`) watches BOTH CachePolicy and CacheTenant and pushes one combined snapshot (avoids the replace-on-write race two reconcilers would cause). The CacheIndex poller adds a per-CacheTenant status projection (`indexEntries`, `Ready`/`QuotaExceeded` conditions, `observedGeneration`) matched by `spec.tenantID`.
- **Tests**: index eviction (table-driven, per-tenant scoped, fail-open), aggregate invariant, policy snapshot round-trip + version bump, reconciler flattening, status condition flap, and an envtest integration covering the full loop + the no-CacheTenant invariant.

## Test plan
- [x] `make pre-pr` (vet, lint, race tests, drift, coverage ≥ 79%)
- [x] envtest integration (`KUBEBUILDER_ASSETS=$(make test-env | tail -1)`): under/over-budget eviction, status reflects post-eviction count, QuotaExceeded flaps True→False, delete reverts to unrestricted

## Backward compatibility
No CacheTenant in the cluster → ingest is byte-identical to today (fail-open).

**Intentional v1alpha1 trim**: `spec.quota.maxMemoryBytes` and `status.memoryUsed` are removed. They shipped with the CRD but were never enforced or populated; per the enforcement-boundary decision, a `max*` field is only added when the cache plane authoritatively owns the resource. Per-tenant byte isolation is an engine/runtime concern (separate engine Deployments + pod memory limits) and is tracked separately.

**Semantic refinement**: `CacheIndex.status.prefixes.summary.total` now reports the total index-entry count (replica×prefix records) — the same unit `maxIndexEntries` bounds and equal to Σ per-tenant `indexEntries`. Identical to the distinct-prefix count except when a prefix is held by multiple replicas.

## Coordination
- The secure-`/policy` work, when it lands, automatically protects this extended snapshot under the same auth (no extra work here) — confirm merge order on adopt.
- The CacheIndex poller now has two status projections (cluster-wide CacheIndex + per-CacheTenant). If a sibling per-CacheBackend projection lands, consider splitting projections into parallel goroutines (future, not this PR).
- Rolling-update note: the version bump means an old controller's v1 push is rejected by the new server until the new controller rolls out; soft-state, self-heals on the next re-push.


EdHasNoLife:
Thanks — addressing the Codex review:

**Blocking — removed `spec.quota.maxMemoryBytes` / `status.memoryUsed` (won't revert):** this is a deliberate, maintainer-approved trim, not an oversight. The cache plane only exposes a `max*` quota for a resource it authoritatively owns. Engine KV memory is a shared, tenant-unaware pool (vLLM/LMCache key by block hash, not tenant), so we can neither enforce a per-tenant byte budget nor honestly observe one — `ReplicaStats.cache_memory_bytes` is the engine total and would be double-counted across tenants sharing an engine. The fields shipped with the CRD but were never enforced or populated, so removing them at `v1alpha1` (pre-release) drops dead schema rather than breaking a working contract. Per-tenant byte isolation is an engine/runtime concern (separate engine Deployments + pod memory limits) and is tracked separately. Rationale is now documented in `docs/design/policy-crds.md`.

**Should-fix — design docs:** fixed. `docs/design/policy-crds.md` and `docs/design/policy-propagation.md` now reflect the removed memory fields and document the `/policy` v2 tenant-quota propagation (`[]ResolvedTenant`, version bump, one reconciler watching both CRs, ingest-time Fairness eviction).

EdHasNoLife:
Latest round (`4ae7cb5`) — addressed the Should-fix items:

- **`cacheindex_controller.go` (absent tenant → nil/Ready=False):** fixed. `reconcileTenantStatuses` only runs after a *successful* scrape, so a tenant with no snapshot row now reports `status.indexEntries=0` (observed zero) with `Ready=True`. Idle, drained-to-zero, and zero-budget tenants surface live usage as `0`; `nil` now means only "no successful scrape yet". Test added.
- **`policy-propagation.md` wire schema:** updated the example to v2 (`"version": 2` + the `tenants[]` slice and per-field docs).
- **`index.go` `Snapshot` comment:** corrected — `TotalPrefixes` counts distinct prefix keys (a prefix on N replicas counts once), matching `Aggregate()` and the tests.

**Blocking (`maxMemoryBytes`/`memoryUsed` removal):** unchanged — deliberate, maintainer-approved enforcement-boundary trim (fields were never enforced/populated; rationale in `docs/design/policy-crds.md`). See prior reply.

EdHasNoLife:
**Disposition of the Codex review** (final):

**Should-fix — all three addressed** (commit `27d38fe`, builds on `4ae7cb5`):
- `cacheindex_controller.go` — a tenant absent from a *successful* `/snapshot` now reports `status.indexEntries=0` (observed-zero) with `Ready=True`, not `nil`/`NotObserved`. Covers idle, drained-to-zero, and zero-budget tenants. Unit test added.
- `docs/design/policy-propagation.md` — wire-schema example updated to v2 (`"version": 2` + the `tenants[]` slice and its field docs).
- `pkg/index/index.go` — `Snapshot` comment corrected: `TotalPrefixes` counts distinct prefix keys (a prefix on N replicas counts once).

**Blocking — acknowledged, and we are keeping the removal (won't change).** The finding is correct that dropping `spec.quota.maxMemoryBytes` / `status.memoryUsed` removes served v1alpha1 fields. This is a deliberate, maintainer-confirmed decision, not an oversight:
- The fields shipped in #29 but were never enforced or populated — no working behavior depends on them.
- We can neither enforce a per-tenant byte budget nor honestly observe one: the engine KV cache is a shared, tenant-unaware pool, and `ReplicaStats.cache_memory_bytes` is the engine total (double-counted across tenants sharing an engine). A `max*` quota is only surfaced for a resource the cache plane authoritatively owns — the index entry table (`maxIndexEntries`), which this PR does enforce.
- The API is `v1alpha1`; per Kubernetes alpha convention, fields may change/remove between iterations without a deprecation cycle. Pruning happens silently (structural schema), so existing manifests still apply.
- Rationale is documented in `docs/design/policy-crds.md`; per-tenant byte isolation (an engine/runtime concern) is tracked separately.

Thanks for the thorough pass — the observed-zero fix in particular was a real gap.

heymrbox (approved):

EdHasNoLife:
Addressed the latest review in `5318664`:

**Should-fix**
- **Duplicate `spec.tenantID` inconsistency** (`cacheindex_controller.go` / `controlplane_controller.go`): the status writer now resolves the *same* effective owner the `/policy` pusher uses (first CR by namespace/name per tenantID). A shadowed duplicate is marked `Ready=False/DuplicateTenantID` + `QuotaExceeded=False/NotEffective` instead of advertising its own non-enforced budget. Unit test `TestReconcileTenantStatusesShadowedDuplicate` added.
- **`PolicyStore.Replace` footgun** (`policy.go`): now delegates to `ReplaceSnapshot(policies, nil)`, so it clears the tenant table rather than leaving stale quotas. No caller can desync the two maps anymore.

**Nit**
- Ready message no longer claims the tenant row existed (absent tenants are synthesized as observed-zero) — it now describes the snapshot read.
- `default_install_smoke.sh` header checklist renumbered (the two `4.`s after the CacheTenant assertion landed → 1–7).

**Blocking (`maxMemoryBytes` / `memoryUsed` removal):** unchanged — deliberate, maintainer-confirmed enforcement-boundary trim (can't enforce or honestly observe per-tenant memory on shared engines; rationale in `docs/design/policy-crds.md`). Per the prior decision on this PR.

EdHasNoLife:
Addressed the new round in `77c83e4`:

**Should-fix**
- **Undocumented metric** (`metrics.go`): `inferencecache_tenant_evictions_total{tenant_id,reason}` is now in the public metrics reference (`docs/reference/metrics.md`), with a `/metrics`-output test (`TestTenantEvictionsMetricExposed`) asserting it's exposed lazily with the documented labels.
- **Negative budget** (`policy.go`): `ReplaceSnapshot` now clamps a negative `maxIndexEntries` to `0` at the `/policy` trust boundary, so a hand-crafted negative budget can't read as "no enforcement" in the index. The CRD already blocks negatives on the legit path; this hardens the direct-POST path. Test added.

**Nit**
- `maxIndexEntries` / `status.indexEntries` godoc now say "distinct prefixes" (a multi-replica prefix counts once), matching enforcement and the design docs. CRD regenerated.

**Blocking (`maxMemoryBytes`/`memoryUsed`):** unchanged — standing maintainer-confirmed enforcement-boundary decision.

EdHasNoLife:
Addressed the latest round in `f848d15`:

**Should-fix**
- **`_default` sentinel collision** (`index.go`): "_default" is now a reserved tenantID — a CEL admission rule (`self != '_default'`) on `spec.tenantID` rejects it, so no real CacheTenant can claim the cluster-aggregate's untenanted-traffic bucket and diverge status from enforcement. Envtest `TestIntegrationCacheTenantRejectsReservedTenantID` verifies the rejection against a real apiserver.
- **No-quota duplicate** (`cacheindex_controller.go`): a duplicate CR with the same `spec.tenantID` but no quota of its own is now also marked `Ready=False/DuplicateTenantID` (was only flagging quota-bearing dups), so it no longer reports `Ready=True/NoQuota` while another CR owns the enforced tenantID. Test extended.

**Nit**
- Status patch error now includes the namespace (`client.ObjectKeyFromObject`).

**Blocking (`maxMemoryBytes`/`memoryUsed`):** unchanged standing decision.

EdHasNoLife:
Addressed the latest round in `c0d7927`:

**Blocking — CEL validation tightening (`cachetenant_types.go:27`):** good catch — fixed by removing the CEL rule entirely. Instead of *reserving* `_default` (a breaking validation tightening on served v1alpha1), the untenanted-traffic bucket now uses the **empty string**, which is collision-free by construction: a real `tenantID` is `MinLength=1`, so no CacheTenant can ever claim the untenanted bucket, and there's nothing to reject. Non-breaking. Added an envtest (`TestIntegrationCacheIndexAcceptsUntenantedTenantRow`) proving the apiserver accepts the empty-id `CacheIndex.status.tenants[]` row (empty `listMapKey`).

**Should-fix**
- `policy-propagation.md`: documented the duplicate-`tenantID` tie-break (first quota-bearing CR by namespace/name wins; shadowed dups go `Ready=False/DuplicateTenantID`).
- `policy-crds.md` reserved-value doc: moot now — there's no reserved tenantID anymore.

**Nit**
- `metrics.md`: `inferencecache_tenant_evictions_total` added to the metric-ownership section; `/policy` description now notes the combined CachePolicy + CacheTenant snapshot.

**Blocking — `maxMemoryBytes`/`memoryUsed` removal:** unchanged standing maintainer decision.

EdHasNoLife:
Addressed the latest round in `a16ad8e`:

**Should-fix**
- **`indexEntries` not on the CacheIndex CR** (`cacheindex_controller.go`): added `indexEntries` to `CacheIndex.status.tenants[]` (additive/optional), so the `Σ tenants[].indexEntries == prefixes.summary.total` invariant is now verifiable straight from `kubectl get cacheindex -o yaml` — not only on `/snapshot` and `CacheTenant.status`. Unit test + envtest assert the equality on the CR.

**Nit**
- `metrics.go`: `serverMetrics` doc refreshed to list the metrics actually registered (was stale at "only `server_up`").
- `policy.go`: reworded the version-bump comment — the version field is the forward-looking guard; it no longer implies a decode-order behavior the code can't guarantee (a v1 server rejects `tenants` at decode under `DisallowUnknownFields`).
- `policy-propagation.md`: the FULL-snapshot summary now notes it carries both policies and quota-bearing tenants.

**Blocking (`maxMemoryBytes`/`memoryUsed`):** unchanged standing decision.

PR #57, opened 2026-05-31
Remove three inert CacheBackend spec/status fields at v1alpha1

EdHasNoLife:
Remove three inert fields from `CacheBackend` (v1alpha1 schema trim; no production users). Each is accepted by the apiserver and persisted, but never read by any code — a silent no-op for operators.

- **`spec.integration.lookupTimeoutMs`** — inert. Duplicates `CachePolicy.spec.lookupTimeoutMs`; only the CachePolicy version is flattened into the server's `ResolvedPolicy` and honored on the `LookupRoute` path. Setting it here does nothing. Defaulter stamp removed.
- **`spec.integration.minimumPrefixTokens`** — inert. Same story: duplicates the wired `CachePolicy.spec.minimumPrefixTokens`. Defaulter stamp removed.
- **`status.indexEntries`** — inert. Status field with zero writers; duplicates `status.indexParticipation.prefixCount`, which is actually written (by the CacheIndex poller) and shown by `kubectl get cachebackend`.

CRD manifest + deepcopy regenerated (drift-clean); samples and `docs/design/cachebackend-api.md` updated; schema test pins the three fields as absent and validates `status.indexParticipation.prefixCount`.

**Why remove, not wire:** wiring the two `integration.*` fields would create a competing source of truth against CachePolicy and break the per-namespace policy abstraction, for no operator benefit; `indexEntries` already has an authoritative replacement.

**Operator note:** if you set these on `CacheBackend`, they did nothing. The authoritative per-namespace surface is `CachePolicy.spec.lookupTimeoutMs` / `CachePolicy.spec.minimumPrefixTokens` (wired into `LookupRoute`); live index counts are on `status.indexParticipation.prefixCount`.

**Test plan**
- [x] `make ci` (naming + fmt + vet + race tests + build)
- [x] `make pre-pr` (drift-clean + coverage gate)
- [x] Defaulter tests updated — no longer assert on removed fields
- [x] Schema test asserts retired fields absent + pins `indexParticipation.prefixCount`
- [ ] envtest: trimmed shape applies; populating a removed field fails apiserver schema validation (skipped locally — `KUBEBUILDER_ASSETS` unset)

**Coordination**
- Same trim pattern as the sibling CachePolicy / CacheTenant field trims (ship in consistent order).
- Sample cleanup overlaps the samples-fix ticket; if that lands first, that portion is a no-op.

PR #60, opened 2026-05-31
Remove CacheBackendStatus.Health; conditions are authoritative

EdHasNoLife:
## Summary

CacheBackend.status carried two parallel readiness signals — the older single-enum `Health` field (`Pending|Ready|Degraded|Failed`) and standard K8s `status.conditions[Ready|Progressing]`. Both were wired by the C2 reconciler with no enforced invariant, and the `kubectl` printer column read `.status.health`. Drops the duplicate Health surface in favour of conditions.

## Why — K8s convention

- Every modern core controller (Deployment, StatefulSet, Pod, ReplicaSet, HPA, Job, …) expresses readiness through `status.conditions`.
- Standard tooling (`kubectl wait`, kstatus) operates on conditions, not on per-CRD enum fields.
- Composable: `Ready=False, Reason=ReplicasUnavailable` is a finer signal than any single Health value.
- Single source of truth — no consistency invariant to maintain between two fields.

## What changed

API (v1alpha1 schema-breaking, intentional):
- Remove `CacheBackendStatus.Health` + the `CacheBackendHealth` enum type and all four constants.
- Replace the `Health` printer column with a `Ready` column using `.status.conditions[?(@.type=="Ready")].status`.

Reconciler:
- `reconcileExternal` / `reconcileUnmanaged` / `updateManagedStatus` drop every `Status.Health` write; the equivalent state was already being emitted on the Ready condition alongside.
- `managedHealth → managedReadiness` (returns only Ready's status/reason/message); `progressingFromHealth → progressingFromReady` (switches on `(readyStatus, reason)`).
- New `conditionReason{BackendReady,ScaledToZero,RolloutInProgress,ReplicasUnavailable}` constants so transition predicates and the Progressing derivation switch on stable strings.
- `stateSnapshot` carries derived `ready` / `degraded` booleans (new `isReady` / `isDegraded` predicates); transition events fire on those predicates with identical semantics.

Tests / docs / canaries:
- Migrate every `Status.Health` assertion across controller + `webhook/pod` tests to the Ready condition.
- CRD schema test positively guards against `status.health` reappearing.
- `canary_c2_reconcile.sh` + `canary_c6_engine_wiring.sh` switch jsonpath to `.status.conditions[Ready].status`.
- Reference-stack README + c2-reconciler-canary workflow narrative updated.

No new condition types added. The existing Progressing condition is preserved.

## Test plan

- [x] `make ci` (race tests + naming + vet + lint + build) — green
- [x] `make pre-pr` — no generated-code drift
- [x] envtest — full CacheBackend lifecycle (apply → Ready=True → degrade → Ready=False/ReplicasUnavailable) asserts via conditions only
- [x] Unit — `managedReadiness` + `progressingFromReady` exhaustive cases


EdHasNoLife:
Addressed in 19a1a4b:

- **Should-fix** — `docs/design/cachebackend-api.md` synced: dropped the `status.health` row, replaced the two-row Ready/Progressing summary with the explicit `(Type, Status, Reason, Meaning)` table grounded in the new `conditionReason*` constants, added the Ready column to the `kubectl get cachebackend` description, reworded the transition-event entry to name the `Ready=False/ReplicasUnavailable` shape directly.
- **Should-fix** — `TestCacheBackendCRDPrintColumns` now asserts the Ready column with its exact JSONPath AND asserts that a Health column is NOT present, so a future regen that re-introduces it fails the test.
- **Nits** — all comment / helper-name references to `managedHealth`, "backend health", "derives Health", "frozen Health", "publishes Health" rewritten to refer to the Ready condition / `managedReadiness`. Renamed `TestManagedHealthIgnoresSpecReplicasUnderHPA` → `TestManagedReadinessIgnoresSpecReplicasUnderHPA`.

**Blocking — v1alpha1 schema break, intentionally kept.** Removing `status.health` is the scope of the ticket, not a side effect: it's a per-ticket waiver of the general "v1alpha1 backward-compatible" rule. Pre-launch state, no current consumers; `Health` is a K8s anti-pattern that earlier v1alpha1 cleanup PRs (CAC-66 / CAC-99 / CAC-101) set the precedent for removing before v1beta1 promotion. The Ready printer column added in this PR replaces Health on the operator-facing display surface, so `kubectl get cachebackend` still shows a readiness signal — just via the K8s-idiomatic conditions path.

EdHasNoLife:
Round-2 addressed in 8264828:

- **Should-fix** — `default_install_smoke.sh` now polls the same `.status.conditions[?(@.type=="Ready")].status` JSONPath the kubebuilder printcolumn annotation writes into the CRD, waiting for `True` after the endpoint check. Operator-facing surface change (printer column) is now smoke-tested on a real install, matching the same shape as the existing endpoint / matchedEnginePods assertions. Catches a CRD-side JSONPath typo that no unit test can see.
- **Nit** — `cachebackend_controller.go` event-reasons godoc + Reconcile apply-error comment + `cachebackend_events_test.go` "drives the health transition" comment now refer to the Ready condition / `Ready=False/ReplicasUnavailable` shape directly.

**Blocking — v1alpha1 break (carried forward from round 1).** Same posture as in [my round-1 reply](https://github.com/cachebox-project/inference-cache/pull/60#issuecomment-4588700208): this is the scope of CAC-102 (per the ticket's locked decision #6), not an accidental break. Pre-launch state, no current consumers of `status.health`, the Ready printer column on the same PR replaces the operator-facing surface, and prior v1alpha1 cleanups (CAC-66 / CAC-99 / CAC-101) set the precedent.

EdHasNoLife:
Round-3 addressed in 9ed3c89:

- **Should-fix** — `default_install_smoke.sh` now parses `kubectl get cb <name>` default output (which renders the CRD's `additionalPrinterColumns`) instead of querying the JSONPath directly. Catches the failure modes Codex flagged: the awk pass `exit 1`'s if no `READY` column header is found, and the subsequent value check rejects an empty cell (which is what kubectl renders when a printcolumn JSONPath is malformed).
- **Nit** — `cachebackend_controller.go:64` event-reasons godoc reflowed; the prior edit broke `\`kubectl describe\`` across a line.

**Blocking — v1alpha1 break (carried forward).** Same posture as in [my round-1](https://github.com/cachebox-project/inference-cache/pull/60#issuecomment-4588700208) / [round-2](https://github.com/cachebox-project/inference-cache/pull/60#issuecomment-4588746138) replies: intentional per CAC-102 locked decision #6; pre-launch, no consumers, Ready column replaces operator-facing surface.

EdHasNoLife:
Round-4 addressed in 69fc50a:

- **Should-fix (docs/design/cachebackend-api.md)** — The Identity section previously stated only the strict "v1alpha1 must remain backward-compatible" rule. Added an explicit "Pre-launch carve-out (active until v1beta1)" paragraph that names the per-change waiver, the three gating conditions (locked design decision + zero current consumers + replacement of operator-facing surface), the closed precedent (`CacheTenant` memory-quota fields), and the current applied removal (`status.health`). The carve-out closes once v1beta1 promotes — subsequent breaking changes require a versioned migration. Documents the policy instead of leaving the doc and the PR contradicting each other.

- **Blocking + Should-fix (status.health removal + schema-test guard)** — Same posture as in [round-1](https://github.com/cachebox-project/inference-cache/pull/60#issuecomment-4588700208) / [round-2](https://github.com/cachebox-project/inference-cache/pull/60#issuecomment-4588746138) / [round-3](https://github.com/cachebox-project/inference-cache/pull/60#issuecomment-4588819795) replies: this is the scope of the locked decision, now documented in the design doc alongside the precedent. The schema-test guard pins the agreed-upon shape.

No further substantive action open from my side. Holding the line on the locked decision.

EdHasNoLife:
Round-5 addressed in f2a7baf — five Nits scrubbed:

- `cachebackend_matched_pods_test.go` (4 sites): four "the status fields written by dispatch" comments used in interceptor-discriminator narration had stale `Health` entries → replaced with `Conditions`. The comments are load-bearing for future status-writer additions, so keeping them aligned matters.
- `pkg/adapters/runtime/vllm_lmcache.go`: TCP readiness-probe comment said the probe gates the Ready condition "via managedHealth" → renamed to `managedReadiness`.

These two files arrived via the round-1 merge from main and slipped through the original Health/managedHealth grep sweep.

**Should-fix (docs/design/cachebackend-api.md:19, waiver audit-trail).** Codex asks the carve-out paragraph to link the originating decision and migration evidence. The project's `verify-no-internal-refs` Make target hard-bans internal ticket IDs (e.g. CAC-XX) and `linear.app` URLs in tracked files (see `make verify-no-internal-refs`); the same hook would reject a "see ticket X" link in `docs/design/`. The audit trail therefore lives in the locked design decision off-repo + the commit / PR title pair, not in tracked-file text. The carve-out paragraph already names the gating conditions and the closed precedent (`CacheTenant` memory-quota fields) explicitly so the policy is self-describing within the tracked files; that is the strongest auditable form compatible with the project rule. Not addressing further.

**Blocking (v1alpha1 break).** Same posture across all five rounds — locked decision, now also documented in the design doc.

EdHasNoLife:
Round-8 addressed in dbd1500:

- **Blocking** — `progressingFromReady` (controller.go) now treats `AwaitingFirstKVEvent` as still-converging (`Progressing=True`), matching `RolloutInProgress`. The merge resolution dropped that branch, which made the gate's wait window appear stuck. `TestProgressingFromReadyExhaustive` gains the `AwaitingFirstKVEvent` and `NoKVEventsObserved` cases — the missing AwaitingFirstKVEvent case is what let the regression slip past CI in the merge commit.
- **Should-fix** — Both canary scripts (`canary_c2_reconcile.sh`, `canary_c6_engine_wiring.sh`) annotate the test CacheBackend with `inferencecache.io/require-kv-events: "false"` so they don't deadlock waiting for a first KV event that arrives only after later steps in the script. The gate itself is covered by its own dedicated canary; these two are about the reconciler + webhook wiring.
- **Nit** — Carve-out paragraph in `docs/design/cachebackend-api.md` names `Conditions[Degraded]` as the direct replacement for the old `Degraded` health enum value; `lmcacheBackend` test-fixture godoc switches "Ready / Degraded health" → "Ready / Degraded conditions".

**Blocking — v1alpha1 break (carried forward).** Per the explicit Schema Evolution Policy in `PROJECT_CONTEXT.md` §5 (pre-launch carve-out, active until CAC-107 v1beta1 promotion), v1alpha1 schema-breaking findings on cleanup PRs are not Blocking — they can be waived with a citation to the policy. CAC-102 is the named precedent in the policy itself.

EdHasNoLife:
Round-9: merged latest main (5 commits) + scrubbed the External-CR HEALTH printer check.

Pushed 454378e:
- **install-smoke fix** — `default_install_smoke.sh` External-CR section came in via PR #65 from main and asserted the `HEALTH` printer column. Switched the column-header check from `HEALTH` → `READY` and the row-value check from `Ready` (the old enum value) → `True` (the condition status). This was the actual CI fail Codex flagged.
- **Should-fix** — replaced the remaining "Pending" terminology with the explicit `Ready=False/AwaitingFirstKVEvent` shape in:
  - `cachebackend_types.go` (firstEventTimeout field godoc)
  - `cachebackend_controller.go` (`evaluateKVEventReadiness` godoc + the operator-facing AwaitingFirstKVEvent Event message)
  - `docs/design/cachebackend-api.md` (firstEventTimeout row)
- Merge brought in #57 (inert spec fields dropped — `indexEntries`, `integration.lookupTimeoutMs`, `integration.minimumPrefixTokens`), #65 (External-CR smoke), #66 (CachePolicy push smoke), #61 (/policy auth), #58 (sample manifest fixes). Conflicts in `cachebackend_types_test.go` + the generated CRD YAML; both resolved.

**Blocking — v1alpha1 break (carried forward).** Same waiver per the Schema Evolution Policy in `PROJECT_CONTEXT.md` §5.

EdHasNoLife:
Round-10 — install-smoke fix.

The install-smoke fail this round was a pre-existing bug on main itself (the `2deecc3` merge commit for #57 also failed install-smoke). PR #57 (drop inert spec fields) moved `lookupTimeoutMs` / `minimumPrefixTokens` from `CacheBackend.spec.integration` to `CachePolicy`, but PR #65's smoke section (Assert External CacheBackend) still pulled those JSONPaths and `fail`ed on the missing values.

Pushed 5b59b1f:
- Drop the now-stale defaulter check from the External-CR smoke section; the CRD-trim assertion earlier in the script (lines ~345-350) already guards against accidental re-introduction.
- Drop the same field list from the script's leading comment block.
- Codex round-9 **Nit** — fold the redundant "still converging / stuck/degraded" paragraph into the Progressing row in the design doc (the row already said the same thing).

**Blocking — v1alpha1 break (carried forward).** Same waiver per the Schema Evolution Policy in `PROJECT_CONTEXT.md` §5.

EdHasNoLife:
Round-10 nits + Should-fix addressed in 128f60e:

- **Nit** — Both stale comments in `default_install_smoke.sh` External section (the mutating-webhook-defaults-integration-knobs claim) scrubbed.
- **Should-fix** — `docs/design/cachebackend-api.md` lead-in to the schema-evolution section now forward-references the carve-out instead of asserting v1alpha1 is additive-only and then immediately contradicting itself one line later. Post-v1beta1 the additive rule applies; until then the carve-out controls.

**Blocking — v1alpha1 break (carried forward).** Same waiver per the Schema Evolution Policy in `PROJECT_CONTEXT.md` §5.

EdHasNoLife:
Round-11 carryover (caught after re-reading round-11's review separately from round-12's; round-12 silently dropped these because the lines weren't in the smoke-fix diff, but the code was unchanged so the findings were still live):

- **Nit** — `internal/controller/cachebackend_controller.go`: restored the lost subject on the `emitTransitionEvents` godoc. An earlier merge resolution chopped the lead-in so the comment started with "Conditions[Ready/Degraded], the KV-event readiness gate state, or the …" as a sentence fragment.
- **Should-fix** — `docs/reference-stack/scripts/canary_c2_reconcile.sh`: default `SKIP_TRAFFIC=1`. The traffic block was a colocated-rendering leftover that port-forwards `svc/$CR_NAME` to `:8000` expecting a bundled vLLM surface — the modern split layout exposes only the LMCache server on `:65432` (TCP lm://) with no HTTP /metrics, so the check cannot succeed without a separately wired engine. Default-off makes the reconciler canary correct out of the box; engine-paired runs flip the knob.

**Blocking — v1alpha1 break (carried forward).** Same waiver per the Schema Evolution Policy in `PROJECT_CONTEXT.md` §5.

heymrbox (approved):

EdHasNoLife:
Round-13 docs sync addressed in 43bca0d (all consequences of the SKIP_TRAFFIC=1 flip in e722a3d):

- **Should-fix** — `.github/workflows/c2-reconciler-canary.yml` header drops "engine prefix-cache hit" from the asserted-by-default list; calls out the optional traffic block as `SKIP_TRAFFIC=0` opt-in.
- **Should-fix** — `docs/reference-stack/README.md` canary section: same shape, traffic block described as opt-in with the wiring requirements named.
- **Should-fix** — `canary_c2_reconcile.sh:41` SKIP_TRAFFIC comment: spells out that "set SKIP_TRAFFIC=0 when running an engine alongside" still requires repointing the port-forward at the engine Service, not the cache-server Service.
- **Nit** — `canary_c2_reconcile.sh:6` script header: drop the "CPU vLLM pods become Ready" success line (it's the cache-server Deployment that becomes Available now; no bundled vLLM pod) and frame the traffic block as opt-in with engine-side wiring required.

**Blocking — v1alpha1 break (carried forward).** Same waiver per \`PROJECT_CONTEXT.md\` §5.

PR #66, opened 2026-06-01
Assert CachePolicy push in install smoke

heymrbox:
# Assert CachePolicy push in install smoke

## Summary

Adds default-install smoke coverage for the CachePolicy PUSH side of the controller/server bridge. The smoke now applies a CachePolicy, checks its Eviction/FailOpen printer columns, seeds synthetic prefix metadata through the existing gRPC API, and verifies the pushed minimumPrefixTokens policy changes LookupRoute behavior.

## Linked issues

Closes [CAC-109](https://linear.app/cachebox/issue/CAC-109/install-smoke-assert-the-cachepolicy-push-policy-half-of-the)

- [x] Pre-commit naming guard passed (`make verify-naming`; `make pre-pr` also passed).

### Quality
- [ ] CI is green. Pending remote CI.

N/A — no CRD or proto contract files changed.


heymrbox:
Addressed in 2293975.

The smoke now lets the server stamp the synthetic ReportCacheState update time, so the seeded prefix is fresh enough for the above-threshold LookupRoute check to return PREFIX_MATCH. I also reset the CachePolicy smoke namespace before applying the sample and delete it after the assertion, so KEEP_CLUSTER=1 reruns cannot pass from stale policy state left behind by a previous run.

Verified locally:
- bash -n docs/reference-stack/scripts/default_install_smoke.sh
- git diff --check
- make pre-pr

heymrbox:
Addressed the diagnostic nit in e3a0501.

On policy-adoption timeout, the smoke now prints both grpcurl stderr files in addition to the high/low LookupRoute responses, so transport/proto/reflection failures show up directly in the job log instead of looking like policy non-adoption.

Verified locally:
- bash -n docs/reference-stack/scripts/default_install_smoke.sh
- git diff --check
- make pre-pr

heymrbox:
Addressed the namespace-reset nit in 3d4cd57.

The CachePolicy smoke namespace reset now uses a bounded kubectl delete wait with a 60s timeout and a clear failure message, so a stuck reused namespace cannot hang the smoke indefinitely.

Verified locally:
- bash -n docs/reference-stack/scripts/default_install_smoke.sh
- git diff --check
- make pre-pr

EdHasNoLife (approved):

EdHasNoLife (commented):
## Review — Assert CachePolicy push in install smoke (CAC-109)

Reviewed against PR head `3d4cd57`, the proto contract, the server gate logic, the controller push path, and the CAC-109 DoD. **No Blocking or Should-fix findings.**

**What I verified myself:**

- **gRPC payload field names** — every field in the `ReportCacheState` (`CacheStateUpdate`) and `LookupRoute` (`LookupRouteRequest`) JSON maps to a real proto field in `proto/inferencecache/v1alpha1/inferencecache.proto`. grpcurl rejects unknown fields, so a typo would break the test — there is none.
- **Gate semantics** — `inferencecache_service.go:92` gates on the *request's* `effectivePrefixTokens(req)`, confirming the high=64 / low=1 logic. `PolicyStore.Lookup` is per-namespace with no global fallback and `MinimumPrefixTokens<=0 → no threshold`, so absent the push the low-token lookup would also match. The test genuinely distinguishes pushed-vs-not (no false pass).
- **Keying linchpin** — `resolveOnePolicy` (`controlplane_controller.go:289`) keys the pushed policy by `cp.Namespace`, matched against `tenant_id`. The smoke applies in `POLICY_SMOKE_NS` and looks up with `tenantId=POLICY_SMOKE_NS` — match holds.
- **Ingest** — `ReportCacheState` ingests unconditionally; an unbudgeted tenant is unbounded by design, so the seeded prefix lands and the high lookup `PREFIX_MATCH`es.
- **Reason codes** — exact-match seeded prefix → `PREFIX_MATCH`; gated request short-circuits to `NO_HINT` before the index. Both exact-code assertions are correct.
- **CRD pre-exists** — `Eviction` (enum `LRU;LFU`), `FailOpen`, and both printer columns are already on main; the sample's `eviction: LRU` / `failOpen: true` are valid with no CRD change. kubectl uppercases headers to `EVICTION`/`FAILOPEN`, matching the grep.
- **`log` redirection** — `log` writes to stdout, so the `>&2` added on the grpcurl-fallback log lines correctly keeps captured responses clean.
- **`bash -n` passes** (no syntax errors). Sample yaml is consumed only by the smoke; smoke runs in CI (`.github/workflows/default-install-smoke.yml:69`). I could **not** run shellcheck (not installed in my environment).
- **DoD** — delivers all three CAC-109 checkboxes (printer columns; observable-side-effect adoption with no engine traffic; compatible with CAC-94's `/policy` auth since adoption is proven via gRPC rather than a `/policy` GET). Nothing silently descoped.

**Nit (not a finding):** the printer-row value checks (`grep -Fq "true"`, `grep -Fq "LRU"`) match anywhere in the table rather than the specific cell — defensible for a single-row smoke.

LGTM.

PR #69, opened 2026-06-01
Add make verify-samples CI gate

EdHasNoLife:
## Summary

Adds `make verify-samples`: spins up an in-process envtest apiserver, installs the CRDs from `config/crd/bases/` and the CacheBackend admission webhook configuration from `config/webhook/manifests.yaml`, then runs `kubectl apply --dry-run=server -f <file>` for every YAML under `config/samples/`. Any sample admission would reject on a real cluster (unknown engine value, removed CRD field, unsupported runtime/backend pair, reserved-arg/env conflict) fails the gate with the verbatim admission error pointing at the offending file.

Closes the loop on the SGLang-default bug pattern fixed by #58 — going forward, schema↔sample drift fails the PR instead of shipping.

## What's in the gate

- **Implementation** — `hack/verify-samples/main.go` brings up envtest, registers the CacheBackend defaulter + validator in-process with the shipping adapter registry (same wiring as the controller binary), waits for the webhook listener, writes a kubeconfig, and shells out to the `kubectl` shipped in `KUBEBUILDER_ASSETS`.
- **Iteration** — every `*.yaml` / `*.yml` under `config/samples/` (no allowlist).
- **Opt-out** — a sample whose top-of-file comment block contains `# verify-samples: skip` is reported as `SKIP` and not applied; for intentionally-illustrative samples that can't apply cleanly.
- **Output** — per-file `OK` / `SKIP` / `FAIL <file> + verbatim admission error` lines, then a summary; non-zero exit on any FAIL.

## Wiring

- `make pre-pr` — runs the gate after the existing CI gate + gen-drift check.
- `.github/workflows/ci.yml` — appended a `Verify samples` step to the existing `test` job (envtest binaries are already installed there for race tests; no new infra).

## Self-test (gate fails on a deliberately-broken sample)

Planted `config/samples/_test/bad.yaml` with `spec.integration.engine: bogus`, ran the gate, removed the fixture. Output:

```
verify-samples: 13 sample file(s) under config/samples
  FAIL config/samples/_test/bad.yaml
    The CacheBackend "deliberately-broken" is invalid: spec.integration.engine: Invalid value: "bogus": no runtime adapter supports the (engine="bogus", backend="LMCache") pair; supported pairs in this build: vllm/External, vllm/LMCache
  OK   config/samples/cache_v1alpha1_cachebackend.yaml
  ...
verify-samples: 12 ok, 0 skipped, 1 failed
verify-samples: 1 sample(s) rejected by admission — see FAIL lines above
exit status 1
make: *** [verify-samples] Error 1
```

Then prepended `# verify-samples: skip` and confirmed opt-out works (file reported as `SKIP`, exit 0). Fixture removed before commit; not in this diff.

## Test plan

- [x] `make verify-samples` against current samples → 12 ok, 0 skipped, 0 failed
- [x] Self-test: broken sample → gate fails with operator-readable error pointing at file + admission reason
- [x] Self-test: opt-out marker → file reported as SKIP, exit 0
- [x] `make pre-pr` green (CI gate + gen-drift + verify-samples)
- [x] No regressions in existing CI gates (race tests, lint, build all pass)


EdHasNoLife:
Re: Codex nit on `Makefile:178` (`verify-samples` depends on `manifests`):

Keeping the dependency intentionally. The gate should validate samples against the CRDs+webhook generated from the **current in-tree types**, not against whatever was last regenerated and committed — otherwise a contributor who changed `api/v1alpha1/*` without rerunning `make manifests` would get a false PASS against stale CRDs. The drift concern Codex raises is handled by `make pre-pr`'s separate generated-code drift check, which runs immediately after `verify-samples` and fails the PR if the regenerated tree differs from HEAD. In CI, the `lint-generated-proto` job already runs the drift check before the `test` job invokes `verify-samples`, so the dependency is a no-op there. Standalone, the side effect is intended: a contributor iterating on CRD types wants the gate to revalidate against their latest changes without remembering an extra step.

Per-file `kubectl` timeout (the Should-fix) is addressed in d5465fe.

EdHasNoLife:
Re: Codex Should-fix on `hack/verify-samples/admission_test.go:57` (webhook port collision under `go test ./...`):

I don't think this is a real risk — flagging it for human eyes before pushing a code change.

The test reads `wopts.LocalServingPort` **after** `env.Start()` and passes it to `webhook.NewServer`. `WebhookInstallOptions.LocalServingPort` defaults to 0; envtest assigns a free OS port at `Start()` time via `net.Listen("tcp", "127.0.0.1:0")` and writes the result back into `wopts.LocalServingPort`. So my test IS using a unique per-process port — the one envtest dynamically picked, not a fixed default.

Cross-package concurrency under `make test-race` (`go test -race -count=1 ./...`) runs each package's test binary in a **separate OS process** by default (controlled by `-p`, defaulting to GOMAXPROCS). Each process runs its own envtest, each calls `net.Listen("tcp", "127.0.0.1:0")` independently, and the kernel hands out distinct ports. There's no shared default port between processes.

This matches the existing pattern at `internal/webhook/pod/envtest_integration_test.go:79-87`, which also reads `wopts.LocalServingPort` after `env.Start()` and has not exhibited flakes against the same test job.

Happy to add a comment in `admission_test.go` pointing this out if it'd help future readers — but I'd push back on changing the port-allocation strategy itself.

heymrbox (approved):

PR #65, opened 2026-06-01
Assert External CacheBackend smoke

heymrbox:
Assert External CacheBackend smoke

## Summary

Extends the default install smoke gate to exercise the External `CacheBackend` admission and status surface.

- Applies the committed External CacheBackend sample.
- Asserts mutating webhook defaults for `lookupTimeoutMs=50` and `minimumPrefixTokens=256`.
- Waits for `.status.endpoint`, `.status.observedGeneration`, and `Ready=True`.
- Verifies `kubectl get cb` renders the CacheBackend printer columns.
- Asserts validating webhook rejection for an invalid External backend.
- Adds all CacheBackends YAML to smoke diagnostics.

## Linked issues

Closes CAC-108

### Vendor-neutral naming
- [x] No banned vendor naming in changed core/default surfaces.
- [x] No cloud-specific integration added outside adapters.
- [x] No Oracle/OCI domain or namespace added.
- [x] Naming checks passed via `make verify-naming`.

### Quality
- [ ] `make build` and `make test` not run; script-only change.
- [ ] `make lint` not run; targeted checks run instead.
- [ ] `make manifests generate` not run; no CRD/generated files changed.
- [x] New install-smoke assertions cover the changed operator-facing behavior.
- [x] Operator-facing CacheBackend admission/status/printer-column surface is now asserted by install-smoke.
- [ ] CI is green.

Verification run:
- `bash -n docs/reference-stack/scripts/default_install_smoke.sh`
- `git diff --check -- docs/reference-stack/scripts/default_install_smoke.sh`
- `make verify-naming`
- `make verify-no-internal-refs`

Full local install-smoke was attempted but blocked because Docker was not running: `Cannot connect to the Docker daemon`.

### Contracts
- [x] N/A — no CRD or proto contract changed.
- [x] N/A — no `v1alpha1` API compatibility change.
- [x] N/A — `proto/` was not changed.

PR #74, opened 2026-06-02
install-smoke: assert every config/samples/*.yaml applies cleanly

EdHasNoLife:
Adds a final apply-clean backstop to `docs/reference-stack/scripts/default_install_smoke.sh`.

**What**
- Loops `config/samples/*.yaml`, runs `kubectl apply --dry-run=server` against the live kind install.
- Fails the gate on any rejection, surfacing the offending filename + admission error.
- Runs **last** (after the per-CRD behavioral phases); admission-level only — no engine pods, no traffic.
- Honors the `# verify-samples: skip` opt-out (same as the envtest gate), plus a no-match guard so a future sample-layout change can't silently make the backstop a no-op.

**Why — complement to `make verify-samples`, not a duplicate**
- `make verify-samples` (PR #69) runs the same per-sample admission assertion at **envtest** level (in-process apiserver + CacheBackend webhook).
- This phase is the **live-kind** complement: catches real-cluster-only failures that pass envtest — webhook cert injection (cert-manager caBundle wiring), NetworkPolicy on the authed `/policy` + `/snapshot` endpoints, RBAC for status writes.
- Revives the smoke-test phase lost when PR #64 was closed as a duplicate of PR #58; adapted to current main (`*.yaml` glob, skip opt-out, no-match guard).

**Validation (local)**
- `bash -n` clean; `make ci` + `make pre-pr` green.
- `make verify-samples`: all 12 current samples pass real server-side dry-run admission (12 ok, 0 failed) — confirms the loop reports all-OK on a clean tree.
- Skip-marker parsing unit-tested against the envtest gate's exact case matrix (8/8); loop pass/fail/skip paths verified against the real samples via a kubectl shim (rejected sample → filename + error + non-zero exit; skip marker → SKIP).
- The real-cluster delta (cert injection / NetworkPolicy / RBAC) is exercised by the install-smoke workflow on PR push — a full local image-build smoke was not run.

Sibling to the per-CRD install-smoke assertions (External CB smoke PR #65, CachePolicy push smoke PR #66); same gate, generic apply-clean assertion rather than per-CRD behavioral.


EdHasNoLife:
Addressed the Codex review in 98491bd:

- **Glob (Should-fix, :1430)** — switched from the flat `config/samples/*.yaml` glob to `find config/samples -type f \( -name '*.yaml' -o -name '*.yml' \) | sort`, mirroring `hack/verify-samples`' `listSamples` exactly. The two gates now select the identical sample set, so a future `.yml` or subdirectory sample can't be covered by the envtest gate yet silently skipped here. No behavior change today (all 12 samples are flat `.yaml`).
- **Overclaim (Nit, :1388)** — narrowed the rationale. `--dry-run=server` stops at apiserver admission, so this phase catches CRD + admission-wiring failures envtest masks (webhook reachability / cert-manager caBundle injection, CRDs as installed by `config/default`). It does **not** create CRs, write status, or hit `/policy`+`/snapshot`, so it exercises no NetworkPolicy or status-write RBAC — added an explicit scope note pointing at the per-CRD behavioral phases for those.

- **No committed parser tests (Should-fix, :1413)** — intentionally not adding in-tree unit tests, for consistency with the established pattern: the entire `default_install_smoke.sh` (1300+ lines, 10 prior phases) and its sibling scripts under `docs/reference-stack/scripts/` carry no committed unit tests — they're integration-tested via the `default-install-smoke.yml` CI job, which exercises this new phase end-to-end against a real kind cluster on every PR (green on this PR). I did validate the new logic locally before pushing: the shell `has_skip_marker` parser was checked against `hack/verify-samples`' exact 8-case `TestHasSkipMarker` matrix (8/8 match, including the surrounding-whitespace and `skip-it`-not-honored cases), and the OK/SKIP/FAIL/no-match control flow was exercised against the real samples plus a `.yml`/subdir fixture via a `kubectl` shim (rejected sample → filename + admission error + non-zero exit; skip marker → SKIP; recursion picks up `.yml` and subdirs). Happy to add a committed shell-test harness as a follow-up if the team wants one for this script family generally.

EdHasNoLife:
Latest round (8f20cf6):
- **:1460 (hung-apply diagnosability)** — added `--request-timeout=30s` to the per-sample dry-run apply, mirroring `verify-samples`' `perSampleTimeout`; a stuck apiserver/admission now fails that sample fast and its filename is surfaced via the rejection branch.
- **:1469 (symlink lockstep)** — softened the "exact" wording. `find -type f` skips symlinks while Go's `filepath.Walk` includes symlinked files (and they also diverge on symlinked dirs), so exact parity isn't cheaply achievable; config/samples has no symlinks, so the realistic sets match — noted the caveat rather than overclaim.
- **:1426 (no committed parser tests)** — re-raise of the earlier Should-fix; intentionally not addressed in-tree, see the [prior rationale](https://github.com/cachebox-project/inference-cache/pull/74#issuecomment-4605774477): consistent with the whole `default_install_smoke.sh` family being integration-tested via the `default-install-smoke.yml` CI job (green here, exercises this phase end-to-end), plus local validation against `hack/verify-samples`' exact 8-case skip-marker matrix and shim-verified pass/fail/skip/recursion paths. Can add a committed shell-test harness as a follow-up if the team wants one for this script family.

heymrbox (approved):

EdHasNoLife:
Round on a81cb95 (after merging main / the gRPC-TLS overlay phase):
- **:1600 (Should-fix — partial `find` walk)** — fixed. The sample list was fed via process substitution (`done < <(find ... | sort)`), whose exit status is discarded, so `pipefail` couldn't see a `find` failure and a walk that errored after emitting some files could slip past the zero-match guard as partial coverage. Now materialized first with an explicit check — `sample_list="$(find ... | sort)" || fail …` — and iterated via a here-string. Verified: a `find` that errors mid-walk now fails the gate *before* the loop; empty dir still trips the zero-match guard; clean tree unchanged (12/12).
- **:1600 (Nit — symlink lockstep)** — unchanged, waived as before: `find -type f` and Go's `filepath.Walk` don't have cheap exact parity for symlinks (they also diverge on symlinked dirs), and config/samples holds only regular files. Documented as an in-code caveat. Can switch to explicitly rejecting symlinked samples as a follow-up if the team prefers a loud failure over the documented caveat.

(Conflict with #68 was resolved by merge: TLS overlay = assertion #11/phase, sample apply-clean backstop = #12 and runs last per design; single combined PASS line.)

PR #75, opened 2026-06-02
Add admission webhooks for CachePolicy and CacheTenant

EdHasNoLife:
Both CRDs shipped with structural CRD (kubebuilder) validation only — no admission webhook. This adds defaulting + validating webhooks mirroring the existing CacheBackend webhook shape, closing the operator-facing admission gap.

## Enforced (hard-reject, not warn)
- **CachePolicy: at most one per namespace.** The reconciler flattens to one ResolvedPolicy per namespace and picks deterministically, so a second CR silently loses intent. Rejected at CREATE, naming the existing policy. UPDATE on the single CR is unaffected.
- **CachePolicy: `evictionTTL > 0` when set.** kubebuilder Duration markers don't constrain sign; a non-positive TTL would be silently clamped to the index default, hiding the typo.
- **CacheTenant: `tenantID` unique within a namespace.** Duplicate tenantIDs collide in the index's `(tenant, model, scheme, hash)` keying. Namespace-scoped by design — tenantID is the operator's identity, not a global UID, so reuse across namespaces is allowed.

Defaulters are no-op stubs — kubebuilder markers own the literal defaults (eviction=LRU, isolationMode=Fairness); they register the mutating-webhook seam for future cluster-state defaults. Cross-CR checks read siblings via the manager's live APIReader and fail closed on a list error.

## Why it's worth the admission cost
At runtime these misconfigurations are invisible: a dropped policy (silent intent loss) or a structural index collision. At admission they're cheap to catch and reported with a clear, actionable message in `kubectl apply` output.

## Validation
- Unit tests (table-driven, fake client): every rule, the v1alpha1 update-tightening seam (only newly-introduced violations rejected), self-exclusion, and fail-closed-on-list-error.
- envtest integration: 2nd CachePolicy in a namespace rejected; duplicate tenantID rejected; both allowed in a different namespace.
- `make ci` + `make pre-pr` green; verify-samples 12 ok; logic coverage 88% (gate 85%).

## Notes
- Schema is stable: validates against the trimmed CachePolicy (failOpen/tenantScoped removed) and CacheTenant (maxMemoryBytes removed) shapes that just landed.
- Sibling to the CacheBackend reserved-args admission rule — same shape (ordered pluggable rule list, hard-reject). No coordination needed with any open PR.


EdHasNoLife:
Thanks for the review — addressed in 09d4ce6.

**Blocking — cross-namespace tenantID shadowing.** The fact is correct: the index/server key tenants by the bare `tenantID` string (namespace-blind), so the same `tenantID` in two namespaces resolves to one quota/index slot. But this is handled deliberately and is **not silent**: the control-plane reconciler dedups deterministically (`resolveTenants`, first by `(namespace, name)`), and the CacheIndex status writer (`effectiveTenantOwners`) stamps `Ready=False/DuplicateTenantID` + `QuotaExceeded=False/NotEffective` on the shadowed CR, naming the effective owner. Cross-namespace reuse is intentionally permitted (it can be deliberate, e.g. a migration) and left to that runtime signal; the same-namespace collision — an unambiguous mistake — is what the webhook hard-rejects. Decision confirmed to keep the webhook namespace-scoped; documented the namespace-blind keying + runtime backstop in the validator doc and `policy-propagation.md` so the contract is no longer implicit.

**Should-fix — docs not synced.** Fixed. `policy-propagation.md` and the `resolvePolicies`/`resolveTenants` comments previously said the CRD doesn't enforce a singleton and called the webhook a future addition; they now describe the shipped best-effort admission + the controller as the authoritative deterministic backstop.

**Should-fix — list-then-admit isn't atomic (both CRDs).** Correct — it's a TOCTOU window under concurrent CREATEs. Documented as **best-effort** in both validators and the design doc; the controller's deterministic dedup is the authoritative invariant, the webhook is fast feedback on the common sequential mistake. Not making it a hard guarantee (that would need a different mechanism, e.g. a singleton lease) by design.

**Nits — stale verify-samples comments.** Fixed all three; they now state the gate registers all three CRD webhooks.

EdHasNoLife:
Round 2 addressed in 329ec09:

- **nil Reader now fails closed** (both validators) — returns a plain config error instead of admitting; a miswired hard-reject rule surfaces loudly. Updated the nil-Reader unit tests accordingly.
- **Deterministic rejection text** — both checks now name the lexicographically smallest conflicting CR (for CachePolicy, the same one `resolvePolicies` treats as effective), independent of apiserver list order. Added multi-sibling unit tests.
- **envtest installs the shipped `config/webhook/manifests.yaml`** now (not a hand-written copy), so it guards the generated webhook wiring against drift; the test only creates Namespaces/CachePolicies/CacheTenants, so the other webhooks' handler paths are never hit.
- **Docs** — documented the tenantID UPDATE-tightening; replaced the stale "webhook validation out of scope" bullet with one that scopes out only field-level validation and points at the cross-CR rules documented above.

EdHasNoLife:
Round 3 addressed in e079705:

- **Install-smoke gate extended** (the CONTRIBUTING.md requirement for operator-facing surfaces). `default_install_smoke.sh` now asserts, on the real kind install, that a **second CachePolicy** in the namespace and a **duplicate-tenantID CacheTenant** are both rejected at admission — exercising the bundle's webhook Service + cert-manager CA-injection path, not just envtest handler logic.
- **evictionTTL doc** synced: noted that the CachePolicy webhook rejects a non-positive `spec.evictionTTL` at admission; the wire-level `<=0` fallback is retained for unset fields and legacy/raced data.
- **Nit** fixed: the envtest opening comment now matches the shipped-manifest install (it previously said "ONLY the CachePolicy + CacheTenant configurations").

heymrbox (approved):

PR #76, opened 2026-06-02
pkg/index: implement LFU eviction; wire CachePolicy.spec.eviction end-to-end

EdHasNoLife:
Implement LFU eviction in `pkg/index`; wire `CachePolicy.spec.eviction` end-to-end.

## What changed
- Restore the LFU enum value; widen the CRD enum to `LRU;LFU` (default stays `LRU`).
- Per-entry `atomic.Int64` access counter (the `prefixes` inner map becomes pointer-valued so the atomic is never copied). Bumped lock-free on lookup HITs that contribute matched tokens to the returned hint set — only in LFU namespaces; LRU lookups skip the bump.
- Cap-based sweep dispatches per-namespace via a unified `(effectiveCount ASC, lastSeen ASC)` sort key (effectiveCount = access count for LFU entries, 0 for LRU), with a deterministic identity tie-break. All-LRU degenerates exactly to the prior oldest-by-`lastSeen` behavior.
- `ResolvedPolicy.Eviction` (lowercase canonical form) + index `EvictionResolver`; `PolicyPropagationVersion` 2→3; reconciler flattens `spec.eviction`.
- New metric `inferencecache_index_evictions_total{algorithm, reason}` (reason ∈ cap|ttl).
- Docs: policy-crds, policy-propagation (v3 schema), metrics reference; sample manifest note.

## No decay (design note)
Counters do not age. The TTL sweep handles staleness regardless of algorithm; LFU only triggers cap-based evict. This avoids the pure-LFU "hot-but-stale entries pinned forever" pathology by leaning on the TTL we already have — no decay loop, no extra goroutine.

## Enforcement boundary
This PR ships in-index eviction (the cache plane's authority); engine-side eviction is engine-owned and tracked separately as a cross-project effort. No proto change, no engine-side change. The CacheTenant quota-eviction path stays separate from this cap-based path.

## Test plan
- Unit (`pkg/index`): LFU lowest-count-first + lastSeen tie-break; counter atomicity under concurrent lookups (-race); policy-switch picks different victims on identical state; chain-lookup bumps every block in a matched run; counter not bumped under LRU; counter survives re-ingest; eviction-metric labels (cap/ttl × lru/lfu).
- Unit (`pkg/server`): snapshot round-trip carries `Eviction`; propagation version pinned at 3.
- Unit (controller): reconciler flattens `spec.eviction` (LFU/LRU/unset→lru).
- envtest: `CachePolicy{eviction: LFU}` drives LFU victims over-cap; `LRU` drives oldest-first victims.
- `make pre-pr` clean (no drift, coverage 86.9% ≥ 85%); `make verify-samples` 12/12.


EdHasNoLife:
**Re: the remaining Codex "Blocking" — `LookupRoute` crediting LFU access counters on delivered hits.**

This is an intentional, accepted design choice, not an oversight:

- **It's the whole point of LFU.** A frequency-based eviction policy has to observe lookup frequency somewhere; for a cache plane whose only authoritative state is the index, the access counter is that signal. The alternative (no on-lookup signal) means no LFU.
- **It's soft, lock-free, eviction-ordering state — not a functional side-effect.** The counter is an `atomic.Int64` bumped via `LookupResult.CreditHits`. It never changes a lookup's `replica_scores`/`reason_code`, never errors, and never affects fail-open behavior. Losing it (restart, race) only changes which entry a *future* cap-based sweep evicts — soft state, same as the rest of the index.
- **Scoped to delivered hits.** After review feedback, the bump was moved out of the index scan into the handler's delivered-response path (`buildLookupResponse`), so a lookup discarded as `TIMEOUT` credits nothing, and `Lookup()` itself is side-effect-free again. Only a prefix-match hint the caller actually receives counts; `TENANT_HOT`/`NO_HINT` credit nothing.
- **LRU is unaffected.** Non-LFU namespaces never capture or bump — zero added hot-path cost.
- **Contract docs updated to match.** `docs/design/grpc-contract.md` and `docs/design/policy-propagation.md` now carve out this narrow, documented exception to "side-effect-free apart from metrics."

Net: the "lookups are side-effect-free" guarantee is preserved in spirit (no functional/observable change to the response or fail-open contract); the only effect is a soft, lock-free counter that exists solely to make the operator-selected `eviction: LFU` algorithm work. Keeping as-is by design.

(Separately: the red `Vulnerability Scan` was three newly-published go1.26.3 stdlib advisories in pre-existing code paths; bumped the toolchain to go1.26.4 — govulncheck now reports no vulnerabilities.)

heymrbox (approved):

PR #77, opened 2026-06-02
Assert server HTTP smoke surface

heymrbox:
## Summary

Adds default install-smoke coverage for the installed server HTTP surface:
- asserts `/readyz` returns 200 through the server Service
- asserts `/metrics` exposes `inferencecache_server_up 1`
- asserts CacheIndex `kubectl get ci` printer columns render `Prefixes` and `Changed`

## Linked issues

[CAC-110](https://linear.app/cachebox/issue/CAC-110/install-smoke-assert-the-server-http-surface-metrics-inferencecache
)

### Quality
- [x] `make manifests generate` produces **no drift** (generated-code drift check passed via `make pre-pr`; no CRD/API changes).
- [x] CI is green.

- [x] N/A - no CRD or proto contract changed.
- [x] N/A - no `v1alpha1` API contract changed.
- [x] N/A - `proto/` was not changed.


heymrbox:
Thanks, fixed this in dcf3558.

The CacheIndex smoke now verifies the actual `kubectl get ci cluster-default` row instead of only checking the header:

- the row includes `cluster-default`
- `Prefixes` renders as `0` for the no-traffic smoke state
- `Changed` is populated and not `<none>` / `<unknown>`

Verification:

- `git diff --check`
- `make pre-pr`

heymrbox:
Also cleaned up the stale phase references in e3a822c.

Both TLS comments now point back to plaintext phase 6, matching the renumbered smoke overview.

Verification:

- `git diff --check`
- `make pre-pr`

heymrbox:
Fixes [CAC-110](https://linear.app/cachebox/issue/CAC-110/install-smoke-assert-the-server-http-surface-metrics-inferencecache)

heymrbox:
Fixed the vulncheck failure in c026dbb by bumping the repo Go version from 1.26.3 to 1.26.4, which is the fixed standard-library release reported by govulncheck. `CONTRIBUTING.md` now matches the new minimum.

Verification:

- `go mod tidy`
- `make pre-pr`
- `make vulncheck` → no vulnerabilities found

heymrbox:
Synced the image-build/tooling side in a2f25b0.

The Docker builder image now uses `golang:1.26.4`, and the tools module Go directive is also `1.26.4`, so Docker builds no longer start from a Go 1.26.3 toolchain while the module requires 1.26.4.

Verification:

- no remaining `1.26.3` references in tracked source
- `make pre-pr`
- `make vulncheck` → no vulnerabilities found

EdHasNoLife (commented):
Reviewed the new HTTP-surface assertions in `default_install_smoke.sh` against the code they exercise. **No correctness bugs found** — the assertions are internally consistent and match the implementation.

### Verified (by inspection — the kind-based smoke itself was not run locally)
- **Port 8080 is correct.** `/readyz` + `/metrics` are served on `:8080` (`pkg/server/server.go:45,122,130`), distinct from `/snapshot` + `/policy` on `:8081` — so the new `:8080` probe is the right surface, not a port mix-up. The `inference-cache-server` Service exposes 8080 (named `http`, `config/server/server.yaml:68`) and the NetworkPolicy leaves it open, so the port-forward resolves.
- **`inferencecache_server_up 1`** is a label-less gauge (`pkg/server/metrics.go:18,51`) set to 1 *before* the listeners start (`server.go:262`) — no scrape race — and is already covered by `server_test.go`. The `^inferencecache_server_up[[:space:]]+1$` regex matches the exposition format.
- **Printer-column parsing matches.** Column order is NAME, Prefixes, Changed, Age (`cacheindex_types.go:125-127`), so awk `$2`/`$3` line up.
- **`Prefixes == "0"` holds in the observed-zero state.** `PrefixSummary.Total` is `int64` with no `omitempty` and is written explicitly (`cacheindex_controller.go:560`), so the cell renders `0`, not `<none>`.
- **`Changed` is populated.** `.status.lastUpdated` is written together with `observedServer` (`cacheindex_controller.go:563`), which the smoke already waited for at the prior step, so the column is set by the time the check runs.

### Note (non-blocking)
The Go 1.26.3 → 1.26.4 bump (CONTRIBUTING.md, Dockerfile, both `go.mod`) is unrelated to the HTTP-smoke scope — bundled in. Harmless, but cleaner as a separate commit.

_I did not run `go build/vet/test` — the substantive change is a bash script needing a kind cluster; the above is inspection-level, cross-checked against the repo's own `server_test.go`._

EdHasNoLife (approved):

PR #83, opened 2026-06-03
test: assert prompt topology smoke columns

heymrbox:
## Summary

Extends the default install smoke gate to cover the schema-only PromptTemplate and PDTopology surfaces.

The smoke now applies the committed samples, documents that neither CRD has a default-install status writer yet, and asserts the operator-facing `kubectl get pt` / `kubectl get pdt` printer columns render.

## Linked issues

[CAC-111](https://linear.app/cachebox/issue/CAC-111/install-smoke-prompttemplate-pdtopology-coverage-verify-wired-into)

### Quality
- [x] `make build` and `make test` pass locally. Covered by `make pre-pr`.
- [x] `make lint` clean (gofmt + go vet). Covered by `make pre-pr`.
- [x] `make manifests generate` produces **no drift** (generated code committed). Covered by `make pre-pr`.
- [ ] New/changed behavior has unit tests. N/A for this shell smoke-gate assertion; the change is exercised by the install-smoke job.
- [x] Operator-facing change asserted by the install-smoke gate.
- [ ] CI is green. Pending.

- [x] N/A — no CRD/proto contract changed.
- [x] N/A — no `v1alpha1` API surface changed.
- [x] N/A — `proto/` untouched.


heymrbox:
Addressed in f951ce9.

I tightened the smoke-script comment so it no longer says the manager registers CRDs/RBAC. It now distinguishes the manifest-installed CRDs/RBAC from the controller manager runtime wiring: the manager adds the Go types to the scheme, but no PromptTemplate render controller or PDTopology reconciler starts in `config/default` yet.

Verification:
- `git diff --check`

heymrbox:
Addressed in 26baf4c.

The PromptTemplate/PDTopology smoke phase now deletes its temporary namespace after the schema/printer-column assertions pass, matching the cleanup pattern used by the other dedicated smoke namespaces. Failed runs still preserve diagnostics because the cleanup only runs after the assertions succeed.

Verification:
- `git diff --check`

EdHasNoLife (approved):

EdHasNoLife (commented):
Reviewed against the repo's `pr-review` checklist and the CAC-111 Definition of Done.

**What it does:** adds an install-smoke phase that applies the committed PromptTemplate + PDTopology samples against the live install and asserts the operator-facing `kubectl get pt` / `kubectl get pdt` printer columns render; renumbers the phase comment block (6→7 … 13→14) to slot the new phase in at #6; adds diagnostics capture for both CRDs; adds the `PROMPT_TOPOLOGY_SMOKE_NS` tunable.

**Delivers the DoD.** CAC-111 asks to first confirm whether a status writer exists, then assert accordingly. The PR documents that both CRDs are schema-only (manager registers the CRDs/RBAC but no reconciler is started) and takes the ticket's explicit "schema-only → assert sample applies + printer columns render" branch.

**Verified by inspection** (I did not run the kind-based smoke — that needs a live cluster; what I personally ran was `git`/`grep`/`sed` against the repo):

- Sample names match the `get pt/pdt` args (`prompttemplate-sample`, `pdtopology-sample`), and both samples are namespace-less so the `-n $NS apply` lands them in the smoke namespace.
- Short names `pt` / `pdt` exist on the CRDs and are collision-free against the other project CRDs and core resources.
- Asserted printer columns match the CRD definitions exactly — PromptTemplate `Revision`; PDTopology `Prefill` (`.spec.prefillPools[*].name`) and `Decode` (`.spec.decodePools[*].name`) — and the asserted values (`prefill-a`, `decode-a`) are derivable from the sample spec.
- No false-pass: `prefill-a`/`decode-a` surface only via the printer columns, not elsewhere in the `kubectl get` table, so a passing assertion genuinely proves those columns render. Empty output (e.g. a failed get) flows to `fail`, not a silent pass.
- Phase renumbering is complete and self-consistent — no stale phase cross-references remain (the TLS-section and backward-compat back-references were updated to 13/7).
- `collect_diagnostics` uses the correct plurals (`prompttemplates`, `pdtopologies`).
- The "no reconciler" claim is true: only `CacheBackendReconciler`, `ControlPlaneReconciler`, and `EnginePodEventsReconciler` are wired in `cmd/controller/main.go`.

**Findings:** no Blocking or Should-fix issues. One inline note on the PromptTemplate assertion (necessarily weaker than the PDTopology one, since the status column is empty with no controller — inherent and acceptable).

Verdict: looks good.

docs/reference-stack/scripts/default_install_smoke.sh
```diff
@@ -596,6 +606,52 @@ if ! grep -q "already claimed by CacheTenant" <<<"$ct_reject_out"; then
...
+  || fail "timed out waiting for prior PromptTemplate/PDTopology smoke namespace $PROMPT_TOPOLOGY_SMOKE_NS to delete"
+kubectl create namespace "$PROMPT_TOPOLOGY_SMOKE_NS" --dry-run=client -o yaml \
+  | kubectl apply -f - >/dev/null
+kubectl -n "$PROMPT_TOPOLOGY_SMOKE_NS" apply -f config/samples/cache_v1alpha1_prompttemplate.yaml >/dev/null
+kubectl -n "$PROMPT_TOPOLOGY_SMOKE_NS" apply -f config/samples/cache_v1alpha1_pdtopology.yaml >/dev/null
+
+pt_table="$(kubectl -n "$PROMPT_TOPOLOGY_SMOKE_NS" get pt prompttemplate-sample 2>/dev/null || true)"
+pt_header="$(printf '%s\n' "$pt_table" | sed -n '1p')"
+pt_row="$(printf '%s\n' "$pt_table" | sed -n '2p')"
+if ! grep -Eq "(^|[[:space:]])REVISION([[:space:]]|$)" <<<"$pt_header"; then
```
EdHasNoLife, line 628: Note (not a blocker, inherent to a schema-only surface): this assertion checks that the `REVISION` *header* renders and the row carries the sample name, but it can't validate that the column actually surfaces a value — `.status.templateRevision` is empty with no render-controller in the default install. Because the header text comes from the printer column's `name:` field independently of its JSONPath, this check would still pass if `.status.templateRevision` were mistyped in the CRD. That's unavoidable here (there's no status value to assert), so it's fine — just worth being aware the PromptTemplate check is necessarily weaker than the PDTopology one below, which does validate real column data (`prefill-a`/`decode-a`). If/when CAC-111's render-controller lands a status writer, this is the spot to upgrade to a value assertion (wait for `templateRevision`, then assert the cell).

PR #84, opened 2026-06-03
CacheIndex: deprecate+zero status.tenants[].memoryUsed at v1alpha1

EdHasNoLife:
## Summary
- `CacheIndex.status.tenants[].memoryUsed` is **deprecated and zeroed in place**, not removed
- Per-tenant memory double-counts on shared engines (engine KV cache is a tenant-unaware shared pool) — same root cause as the CacheTenant memory trim
- The per-tenant accumulator is dropped, so the field is always `0`; it stays in the CRD schema **and** the internal `/snapshot` wire shape for controller/server skew-compatibility, with a `Deprecated` godoc and removal deferred to v1beta1
- Per-replica `cacheMemoryBytes` (honest, engine-total-per-replica) stays

## Why deprecate instead of remove
Review flagged the original hard removal as a breaking v1alpha1 status-shape change (the field is already published). Pivoted to the K8s-idiomatic deprecate-don't-remove path: shape preserved, value zeroed, scheduled for v1beta1 removal.

## Operator-facing note
Per-tenant memory on a shared engine cannot be honestly attributed. `status.tenants[].memoryUsed` is now always `0` (deprecated). The replica-level `CacheIndex.status.replicas[].cacheMemoryBytes` remains (per-replica engine total) — operators reasoning about memory should read it there, with the understanding that on a tenant-shared engine the value covers ALL tenants on that replica.

## Test plan
- [ ] make pre-pr drift-clean
- [ ] Index no longer accumulates per-tenant memory (snapshot tenant MemoryUsed == 0 despite non-zero replica CacheMemoryBytes)
- [ ] `/snapshot` + CRD still carry the `memoryUsed` key (skew-safe)
- [ ] install-smoke asserts the deprecated field is retained and the per-replica field is present


EdHasNoLife:
Replies to the Codex review on the prior commit:

**Blocking — `cacheindex_types.go:47` v1alpha1 backward-compat.** Waived per `PROJECT_CONTEXT.md` §Schema-evolution policy: v1alpha1 explicitly permits removing inert fields (no runtime behavior, no known consumers) pre-launch. This field has no consumer (no dashboard, no CLI, no Prometheus metric reads it). The same carve-out was applied to CAC-66's removal of `CacheTenant.spec.quota.maxMemoryBytes` and `CacheTenant.status.memoryUsed`, and to CAC-63 / CAC-99 / CAC-101 / CAC-102. The cleanup ticket sits under §Forward-invariant inert-field hygiene and this is its cluster-aggregate symmetry. The policy carve-out expires at v1beta1 (separate tracking ticket); until then, schema-breaking trims of inert v1alpha1 fields are explicitly permitted.

**Should-fix `pkg/index/index.go:1209` — `/snapshot` JSON wire skew.** Same project policy applies. `/snapshot` is the internal controller→server HTTP wire (`pkg/index/index.go` Snapshot encodes / `internal/controller/cacheindex_controller.go` decodes); both binaries ship as one release with kustomized matched-version deploys, so an older controller talking to a newer server is a transient rolling-window only. The decode-as-zero behavior writes zero to a field that v1alpha1 itself just removed from the CRD — no consumer sees the misleading value. Per the same §Schema-evolution policy this is permissible at v1alpha1; post-v1beta1 we'd add a versioned `/snapshot` and a skew-tolerant decode, but that's premature today.

**Should-fix `api/v1alpha1/cachetenant_types.go:68` — stale godoc.** Fixed in 637bd04 — the godoc now repoints to `CacheIndex.status.replicas[].cacheMemoryBytes` (per-replica engine total) instead of the removed cluster-wide tenant aggregate.

**Nit `pkg/index/index_test.go:220` and `:267` — comment / assertion mismatch.** Both fixed in 637bd04 — the comments now describe what's actually asserted (HitRate is replica-deduped per tenant; IndexEntries is not).

EdHasNoLife:
Replies to the Codex review on 7c38450 (round 3):

**Blocking — `cacheindex_types.go:47` v1alpha1 backward-compat.** Same waiver as the prior round: this is permitted by the alpha-removal carve-out in \`docs/design/cachebackend-api.md\` (zero current consumers, the operator-facing surface is replaced by the per-replica \`status.replicas[].cacheMemoryBytes\`). The carve-out expires at v1beta1.

**Should-fix \`docs/concepts/cacheindex-cluster-aggregate.md:25\` — design-doc sync.** Fair point — fixed in f33dcee. The carve-out paragraph in \`docs/design/cachebackend-api.md\` now names this PR's removal in the closed-precedent list, and \`docs/design/crd-contract.md\` §Enforcement boundary now spells out the cluster-aggregate symmetry alongside the existing CacheTenant explanation.

**Should-fix \`pkg/index/index.go:1212\` — \`/snapshot\` wire skew.** Same waiver as before. \`/snapshot\` is the internal controller→server HTTP wire; both binaries ship in one matched-version release. The decode-as-zero behavior writes zero to a CRD field that this same PR removes — no consumer reads the zero. The frozen-wire-contract test catches drift between the two binaries' compiled types; a deliberate field removal that lands in both at the same release is exactly the schema-evolution case the policy permits.

EdHasNoLife:
Addressed the latest review:

- **`cacheindex_controller.go:607` (skew-discard untested)** — added a `buildCacheIndexStatus` unit assertion that feeds a non-zero `TenantSnapshot.MemoryUsed` and asserts the built status zeroes it, so a future regression that copies the deprecated value back is caught.
- **`policy-crds.md` (doc mirror)** — the CacheIndex section now records the deprecated/always-0 `status.tenants[].memoryUsed` compatibility semantics, matching `crd-contract.md`.

On **`cacheindex_controller.go:179` (write-only-on-change may skip backfilling the explicit `memoryUsed: 0` key on a pre-upgrade row)** — intentionally not changing the diff logic. `statusEqual` compares typed `CacheIndexStatus`, where an absent key and an explicit `0` are the same value, so a fully-quiescent pre-upgrade zero-memory row won't rewrite immediately. This is acceptable eventual consistency for a soft-state, deprecated, semantically-zero field: (1) a consumer honoring the deprecation treats absent ≡ 0; (2) the row backfills the explicit key on the next genuine status change, which on a live index happens within the ~25s poll; (3) the realistic upgrade path is *from* the previously-populated non-zero field, where the values differ and the write fires immediately. Making the equality wire-level would require fetching the raw unstructured object every reconcile — reintroducing the exact resourceVersion churn the write-only-on-change discipline exists to prevent. Not a worthwhile trade for a deprecated field slated for v1beta1 removal.

heymrbox (approved):

PR #96, opened 2026-06-06
controller: bound the lmcache-server container by cgroup memory limits

EdHasNoLife:
## Summary

- Add `CacheBackend.spec.resources *corev1.ResourceRequirements` with a kubebuilder default of `{requests: {memory: "4Gi"}, limits: {memory: "8Gi"}}` so every admitted CacheBackend renders the cache-server pod bounded by a cgroup memory limit. The kubelet was OOM-killing the un-limited `lmcache-server` pod within minutes of T2 write load in the Phase-2 cache-stress benchmark — the kubebuilder default stops that on the minimal-YAML path, the field lets operators tune per-deployment, and a new validating-webhook rule (`rejectResourceLimitsBelowRequests`) rejects an inverted `limits[X] < requests[X]` at admission across every resource present in the intersection.
- Renderer change in `pkg/adapters/runtime/vllm_lmcache.go`: `defaultServerResources` now `DeepCopy`'s `cache.Spec.Resources` so the informer-cached object is never mutated, and only fills in a 250m CPU fallback under autoscaling (the HPA-utilization denominator) without overwriting an operator-supplied CPU request.
- Default-install smoke + envtest integration assert BOTH halves of the contract end-to-end against a real apiserver / real kubectl-installed bundle — the persisted CR's `spec.resources.limits.memory` and the rendered Deployment container's `Container.Resources.Limits[memory]` must both equal 8Gi — because either half-failing reintroduces the OOM-kill regression.
- `config/samples/recipe-gpu-production.yaml` demonstrates a per-deployment memory override on top of an autoscaling block, with prose calling out that the renderer's CPU fallback fills in the HPA denominator automatically; minimal samples inherit the CRD-schema default with no changes needed.

## Test plan

- [x] `pkg/adapters/runtime/vllm_lmcache_test.go` — operator override honored, in-place mutation forbidden, autoscaling CPU fallback fills only when missing, autoscaling never overwrites operator CPU, empty `spec.resources: {}` is respected (no fallback memory)
- [x] `internal/webhook/v1alpha1/cachebackend_webhook_test.go` — `limits.memory < requests.memory` rejected, equal admitted, requests-only / limits-only admitted, CPU coverage pinned
- [x] `internal/controller/cachebackend_resources_integration_test.go` — envtest verifies the CRD-schema default is stamped by the apiserver AND the controller threads it into the rendered Deployment; envtest also verifies operator override is honored end-to-end
- [x] `docs/reference-stack/scripts/default_install_smoke.sh` — new assertion against the live `cb` + `deploy` (operator-facing, runs on the kubectl-installed bundle)
- [x] `make verify-samples` admits all CacheBackend samples through the new webhook rule
- [x] `make ci` clean (vendor-naming, no internal tracker IDs, gofmt, vet, golangci-lint, race tests, build)
- [x] `make pre-pr` clean (no generated-code drift, sample admission)


EdHasNoLife:
Round-11 review response:

**Addressed:** `rejectFractionalExtendedResources` lands in this commit. The check is cheap (`Quantity.AsInt64()`), matches K8s' "extended resources are whole units" rule, and produces a clear field-scoped error at `kubectl apply` instead of letting `nvidia.com/gpu: 500m` reach the apiserver's Pod validator. New unit tests pin both the rejection and the no-touch contract on overcommittable resources (250m CPU still admits).

**Declined, with reasoning:**

1. **Requests-only non-overcommittable resources (no matching limit).** I'm not confident K8s actually rejects this shape. The two upstream paths I can find on `Container.Resources` validation enforce `limits == requests` only when both are present; a requests-only extended resource is, in my reading, admitted by the apiserver and the scheduler uses the request as both the demand and the implicit ceiling. Adding admission to reject it would be **stricter than K8s**, not a mirror — operators would see a CR rejection for a shape K8s itself accepts. Until that behavior is reproduced (a simple `kubectl apply` of a Pod with `requests.nvidia.com/gpu: 1` and no limit), I'd rather not pre-emptively narrow.

2. **Hugepage-size alignment** (`hugepages-2Mi` quantity must be a multiple of 2Mi). True K8s rule, but: (a) the alignment is per-page-size and requires knowing the valid page sizes for the cluster's kernel, which isn't a property the validator has; (b) the realistic shape of the cache-server pod doesn't declare hugepages; (c) `rejectInvalidResourceNames` already validates the suffix is a positive Quantity, so the malformed-name class is closed. If a hugepage-using CR slips an unaligned quantity through, the apiserver gives the operator the same clear `must be a multiple of <pagesize>` error a moment later — the failure mode isn't silent.

The validator's promise from the ticket onward has been "if it would fail later in a way that's hard to diagnose, fail at admission." For #1 and #2 above, the alternative failure mode is a loud, well-messaged apiserver rejection on the rendered Pod — not the silent reconciliation hang we were closing. I'd rather land the PR than keep walking the long tail.

Open to being wrong on #1 — if there's a concrete reproducer of K8s rejecting `requests.nvidia.com/gpu: 1` without a limit, I'll add the rule.

heymrbox (approved):

PR #101, opened 2026-06-08
docs: index sizing & memory tuning guide

EdHasNoLife:
## Summary

- Adds `docs/operations/index-sizing.md`, a new operator-facing tuning guide for the in-memory cache-state index. Covers per-entry memory footprint, workload-shape sizing (chatbot, multi-tenant RAG, long-context coding assistant), TTL trade-offs, the metrics-driven tuning workflow, and rationale for the current defaults.
- Adds `hack/index-sizing/`, a synthetic-load harness that ingests N prefix entries through the real `pkg/index` code path and reports heap + RSS. Not built by `make build`; not picked up by `make test` for coverage (excluded via the existing `/hack/` rule in `COVER_EXCLUDE`).
- Cross-references the new guide from `CachePolicy.spec.evictionTTL` and `CacheTenant.spec.quota.maxIndexEntries` godoc; CRD manifests regenerated.

## Measurements

Harness sweep on `go1.26.4` / `darwin/arm64` (single tenant, single model, 32-byte hash, 1 replica/prefix):

| Entries | `heap_alloc` | RSS | bytes / entry (heap) |
|---:|---:|---:|---:|
| 500,000 | 241 MiB | 300 MiB | 504 |
| 1,000,000 | 481 MiB | 541 MiB | 504 |
| 1,500,000 | 641 MiB | 693 MiB | 448 |

Linear above ~100K; ~500 B per distinct prefix key, ~50 B per additional replica holding the same key, ~30–50 MiB fixed Go-runtime baseline.

## Default re-evaluation

Defaults stayed unchanged after the measurements. Rationale documented in the guide:

- `DefaultMaxEntries = 1,000,000` — one-pod fit for a 1 GiB server budget, comfortably above realistic single-tenant load.
- `DefaultTTL = 30m` — matches typical chat-style prefix-reuse windows; `CachePolicy.spec.evictionTTL` is the per-namespace override.
- `DefaultSweepInterval = 1m` — cheap full-walk at 1M entries.

The guide notes one real gap: the global cap / TTL / sweep interval are currently compile-time constants, not server flags. Operators that need more than 1M entries today route through per-namespace `evictionTTL` shortening or per-tenant `maxIndexEntries` tightening. A follow-up to expose the cap as a flag is mentioned in the guide but out of scope here.

## Test plan
- [x] `make ci` green (lint, race tests, build)
- [x] `make pre-pr` green (generated-code drift clean, verify-samples passes)
- [x] Naming check (`make verify-naming`) clean
- [x] Harness exercised at 500K / 1M / 1.5M entries plus a multi-replica and a multi-tenant variant — all numbers in the doc reproducible via `go run ./hack/index-sizing`


heymrbox (approved):

EdHasNoLife:
Round-8 Codex nits — disposition:

- **docs:136 (~62 vs ~63 hashes)** — fixed in d599a9b. `ceil(1000/16) = 63`, not 62.
- **docs:218 (vague "follow-up")** — fixed in d599a9b. Reworded to a neutral "compile-time today, runtime-tunable via the per-namespace and per-tenant CRs" callout and swept the other "follow-up flag" / "until the cap is a flag" mentions to match. Not filing a public issue inline with this PR — the doc now describes the state honestly without promising work it doesn't track.
- **harness:41 (no tests for validation paths)** — **skipping with reason**. The harness is operator tooling in `hack/`, excluded from coverage (`/hack/` in `Makefile` `COVER_EXCLUDE`), and its validation paths are simple arithmetic (positive flag check, divisibility check, multiply-overflow check) where table tests would mostly be testing the test cases themselves. The trade is real but I'd rather lean on `go run`-time observation than carry a test surface on a non-shipping tool. Happy to revisit if the harness grows into a more general benchmarking helper.

PR #104, opened 2026-06-09
feat: affinity routing for diffuse single-turn workloads

EdHasNoLife:
## Summary

- Adds `CachePolicy.spec.affinityRouting` (Enabled/Disabled, default Enabled).
- On the `LookupRoute` NO_HINT path — both the post-ranker StrategyNone branch AND the pre-lookup `minimumPrefixTokens` short-circuit — the server now returns a stable consistent-hash replica assignment with `reason_code=AFFINITY_HINT` so repeat prompts pin to the same replica and warm T1 on diffuse single-turn workloads (chatbot, RAG with distinct corpus chunks per query). Operators measure the share via `inferencecache_lookup_route_calls_total{reason_code="AFFINITY_HINT"}`.
- Seed = SHA-256 over length-prefixed `block_hashes` (length-prefixed so cross-width proto `bytes` from different engines can't silently collide — vLLM is 8B, SGLang may differ); falls back to `prefix_hash` for legacy callers. No seed → fall through to NO_HINT.
- New `Index.AffinityHint(tenant, model, seed)` — sorted by `replica_id`, modulo over `len(replicas)`. Reuses the per-(tenant, model) `replicasByModel` accelerator under RLock.
- `PolicyPropagationVersion` bumps 5 → 6. `normalizePolicySnapshotForVersion` backfills the missing `affinityRouting` pointer with `DefaultAffinityRoutingEnabled` on v3/v4/v5 bodies so a server-first rollout doesn't silently flip every namespace's affinity off.
- gRPC contract: **no schema change**. `reason_code` is a string by design; the new `AFFINITY_HINT` value is documented in `docs/design/grpc-contract.md` (new "Affinity routing" section + reason-code precedence rules) and the proto field comment. `UNKNOWN_TENANT` / `UNKNOWN_MODEL` / `UNKNOWN_HASH_SCHEME` / `TIMEOUT` keep precedence; `AFFINITY_HINT` never preempts a real `PREFIX_MATCH` or `TENANT_HOT`.
- `affinityEligible` mirrors the index's fail-soft input checks: empty `hash_scheme` and chain arrays with mismatched length stay on NO_HINT (gateway misconfiguration shouldn't get papered over by a stable replica pick).

## Test plan

- [x] `make ci` green locally (race tests + naming + fmt + vet + build)
- [x] `make pre-pr` clean (CI gate + generated-code drift + sample verify)
- [x] Unit: `pkg/index/affinity_test.go` — stability across 100 calls, balance over 1000 deterministic-RNG seeds across 8 replicas in [80, 220] band, empty replica set returns ok=false, single-replica deterministic, sort-order independence
- [x] Unit: `pkg/server/affinity_routing_test.go` — all branches (enabled, disabled, no replicas, no seed, stable repeats, zeroed scoring fields, bypasses minimumPrefixTokens, minimumPrefixTokens still filters when affinity disabled)
- [x] Unit: `pkg/server/policy_test.go` — AffinityRoutingEnabled resolver + v5-body normalization backfill
- [x] Unit: `internal/controller/controlplane_controller_test.go` — resolveOnePolicy mapping (omitted / Enabled / Disabled)
- [x] Envtest: `internal/controller/cachepolicy_affinity_routing_integration_test.go` — full CRD → controller → /policy → resolver loop, including kubebuilder default materialization belt-and-braces readback
- [x] Existing `TestLookupRouteChainNoOverlapNeverFallsThroughToTenantHot` updated to disable affinity in its PolicyStore (preserves its named chain-vs-TENANT_HOT invariant; the affinity behavior on the same scenario is covered in the new test suite)
- [ ] **Follow-up (not a merge gate):** re-run `cache-stress-extreme` lookup mode on `dp-us-chicago-1-dev` after this lands; verify routing-affinity rate is high. Coordinate with team.

## Notes

- The version-pin test was renamed `TestPolicyPropagationVersionIsV5` → `...IsV6` and its godoc extended.
- The CRD `affinityRouting: Enabled` line was added to `config/samples/cache_v1alpha1_cachepolicy.yaml` so operators see the knob and the install-smoke exercises the defaulting path.
- Phase 2 follow-ups (not in scope here): Rendezvous / HRW hashing in place of modulo, `LookupPDRoute` affinity, per-tenant affinity weights.


EdHasNoLife (commented):
## Intent-alignment review (does this match the ticket's stated intent?)

A requirements pass against the ticket, not a line-by-line correctness review — happy to do that separately.

### ✅ Core direction matches
- AFFINITY hint on the `NO_HINT` / `StrategyNone` path — the ticket's central idea ("make `NO_HINT` *useful*").
- New `reason_code` (`AFFINITY_HINT`) — the ticket's *preferred* contract option (the explicit-reason-code variant), for metric visibility.
- Single replica, `matched_tokens=0`, no cache-evidence claim.
- Scheme-aware replica set + scheme-excluded seed — matches the ticket's subtle requirement (seed = prompt identity; replica set scoped so a vLLM request can't pin to an SGLang replica).
- Per-namespace toggle, `PolicyPropagationVersion` bump + back-compat normalization, docs, envtest.

### ⚠️ Deviations from *explicit* ticket requirements

**1. Defaults ON; the ticket requires default OFF.** The ticket's design + definition-of-done call for the routing-strategy field defaulted to *off* ("no behavior change for existing operators"). This PR ships `affinityRouting` with `+kubebuilder:default=Enabled`, so every existing deployment silently flips on — the opposite of the opt-in stance.

**2. The mandatory hot-spot / pressure fallback is missing.** The ticket marks this **mandatory**: when the affinity-pinned replica's `pressure` exceeds a per-namespace threshold (`affinityHotSpotPressure`, default 0.85), step *down* to round-robin among lower-pressure replicas — plus a Zipfian regression test. This PR has no pressure step-down; it's naive `SHA-256(seed) % len(replicas)`. On a skewed (Zipfian) workload the hot prompt pins all its traffic to one replica → saturation, the exact failure mode the ticket pre-empts. **Combined with default-ON, this is the pairing I'd worry about most.**

**3. Naive modulo, not consistent hashing.** (acknowledged in the PR) The ticket requires the hash to be *consistent across replica scale-up/down* (virtual nodes / hash ring, so a single pod add/remove shifts only `1/N`). `% len(replicas)` reshuffles ~all prompt→replica mappings on any pod add/remove → mass T1 eviction churn under autoscaling. Deferring HRW is reasonable **if** tracked + noted as a scale-stability gap.

### 🟡 Process / definition-of-done
- **Scope gate:** the ticket blocks shipping on the analytical-model verdict placing this workload in the "routing-is-the-lever" case (vs "capacity is the bottleneck"). That verdict isn't referenced here.
- **Benchmark validation** (DoD: measurable T1 hit-rate lift) is deferred to a post-merge follow-up.
- **TENANT_HOT composition:** the ticket *preferred* "hash-affinity filtered through tenant-warm candidates"; this PR uses "TENANT_HOT wins outright." The latter is a listed option, so acceptable — just not the stated preference.

### 🔴 The big one: the evidence is now stale
The "~5% T1 hit rate" and the cache-stress numbers that justify this lever were measured **with routing broken** (the content-fingerprint bug → 100% `NO_HINT`; fix in flight as #105). That fix changes the baseline fundamentally: recurring-and-resident prompts now get a real `PREFIX_MATCH` and route to their holder *without* affinity. Post-fix, `AFFINITY_HINT` fires only on *true* misses (novel / evicted), and its remaining value is specifically the **capacity-constrained / eviction-heavy** regime (stable partitioning so an evicted-then-recurring prompt re-caches on the same pod).

→ **Re-run cache-stress on the post-#105 baseline (affinity OFF) first**, then with affinity ON, and keep this only if it adds measurable lift over that baseline. That's exactly the "capacity-bound vs routing-is-the-lever" scope-gate the ticket already demands — it just has to run on the corrected baseline.

### Suggested before merge
1. Default → **OFF** (opt-in, per DoD) — or, if defaulting ON, the **hot-spot fallback is non-negotiable** first.
2. Add the pressure step-down + a Zipfian regression test (the ticket's only mandatory safety net).
3. Re-benchmark on the fixed-routing (#105) baseline; confirm the lift.
4. Modulo→HRW as a tracked fast-follow.

**Net:** the idea and wiring are right and well-tested, but as written it diverges from the ticket on the two firm requirements (default-OFF, mandatory hot-spot fallback), and the benchmark that motivated it needs re-running on the corrected baseline before it's clear this lever still pays off.

EdHasNoLife:
Rebased onto current `main` (collapsed the prior docs-iteration commits into one). Three integration decisions made against the evolved hot path — flagging for review:

- **`PolicyPropagationVersion` 6 → 7.** Main's lookup-strategy gates already took v6; this feature's `affinityRouting` field becomes v7 so a mixed-version rollout gets a clean "unsupported version" rejection instead of a same-version body that fails `DisallowUnknownFields` on decode. Normalization fills `affinityRouting` for any body `< 7`.
- **Affinity seeds from the request's `block_hashes` / `prefix_hash`.** It does not yet fire for the token_ids / prompt_text dual-input modes (no explicit chain on the wire → no seed → fails open to `NO_HINT`). Deliberate scope boundary for this rebase; extending the seed to the server-resolved chain is a clean follow-up.
- **Composition with the existing strategy gates.** A sub-`minimumPrefixTokens` request and a tenant-hot-disabled downgrade both land on `StrategyNone`, which the affinity fallback now fills with `AFFINITY_HINT` (default-Enabled). Two strategy-isolation tests and the fingerprint E2E were updated to reflect this (the E2E now asserts `AFFINITY_HINT` with `matched_tokens == 0`, a stronger no-false-prefix-match signal than the old `NO_HINT`).

heymrbox (approved):

PR #105, opened 2026-06-10
feat(index): content-fingerprint routing key (CAC-178)

EdHasNoLife:
## Why
vLLM seeds its block hash with a per-process-random `NONE_HASH` (PYTHONHASHSEED unset) → engine hashes aren't reproducible across replicas or externally → the gateway can't match them → every `LookupRoute` returns `NO_HINT`.

## What
Key the index on our own deterministic **content fingerprint** instead of the engine hash.

- `pkg/fingerprint`: XXH3-64 (seed 1337) content + rolling prefix hash — matches SMG `event_tree.rs`, so both interoperate on one index
- subscriber: capture `token_ids` + `parent_block_hash`, hash **in-pod** — only hashes leave the pod (control plane stays metadata-only)
- positional indexer: chain the prefix hash across incremental store events via a per-replica reverse map (port of SMG `apply_stored`/`removed`/`cleared`)

## Tests
Golden parity vs SMG/Python (byte-for-byte); cross-event chain == full-prompt roll (the ingest↔query match property); reporter + LMCache round-trip integration tests updated; `go test ./...` green.

Unblocks CAC-178 (warm-routing benchmark). Phase 2 (server-side tokenizer + pass-tokens-to-engine) is tracked separately.


EdHasNoLife:
### Note for reviewers: the remaining automated `changes-requested` item is an accepted, documented deferral

Across the review rounds, every actionable finding has been addressed — most notably a real L2-mode memory leak (the reverse map now prunes on `BlockRemoved` even when the eviction is suppressed, with a regression test), decoder coverage for the now load-bearing `token_ids` / `parent_block_hash`, a broadened "produced no index entries" warning on the ingest path, de-vacuuming the clear-supersedes test, removal of the dead engine-hash helpers, `go mod tidy`, and the contract-doc fingerprint spec. CI is green.

The one finding that intentionally remains is **`positional.go` — an unknown non-root parent is indexed as a fresh root**, which could yield a false `PREFIX_MATCH` after a dropped event or a subscriber restart. This is a **bounded soft-state cost**, not a correctness bug:

- A wrong hint **fails open** — it degrades to a cache miss, never a wrong answer.
- It **cannot happen on a clean cold start** (no gaps), which is the regime this change targets.
- The general hardening — learn the engine's per-process root (`NONE_HASH`) and drop *true* gaps rather than treating them as fresh roots — can't be done by simply "dropping unknown parents," because the engine sends `NONE_HASH` (a non-nil value) for genuine roots, so that would drop real roots too and regress to all-`NO_HINT`. The correct fix is tracked as a follow-up and noted inline in `positional.go`.

So the residual `changes-requested` reflects a deliberately deferred hardening, not an oversight — flagging it here so a human reviewer can merge with that context.

heymrbox (approved):

PR #107, opened 2026-06-10
feat: server-side tokenization — dual-input LookupRoute + cgo tokenizer + token-ID engine client

EdHasNoLife:
Phase 2 of unified-router tokenization: the IC server owns tokenization, so the routing fingerprint is computed over the **same tokens the engine caches** — match by construction, no tokenizer-parity dependency.

> Stacked on #105 (content fingerprint). Base is \`cac-178-content-fingerprint\`; retarget to \`main\` + rebase once #105 merges.

**Dual-input \`LookupRoute\`**
- Request gains \`token_ids\` (9) + \`prompt_text\` (10); response gains \`token_ids\` (4, echoed only on the prompt_text path).
- Precedence: explicit chain > token_ids > prompt_text; all funnel into the existing block-chain match. Fail-open throughout (tokenizer unavailable/error → NO_HINT).
- \`pkg/fingerprint.Chain\` mirrors the subscriber's per-block ingest. proto is additive-only; \`grpc-contract.md\` updated in lockstep.

**Server-owned tokenizer** (\`pkg/tokenize\`)
- Interface + \`Unavailable\` default — pure-Go builds need no Rust; the prompt_text path fails open there.
- cgo impl behind \`-tags smgcgo\` links \`rust/ictokenizer\`, a C-ABI shim over SMG's \`llm-tokenizer\` (git dep; no gateway, no Python). \`make tokenize-cgo-test\`; verified against the real Qwen2.5-0.5B-Instruct tokenizer.

**Token-ID engine client** (\`pkg/engineclient\`)
- OpenAI \`/v1/completions\` with a token-ID prompt (stock vLLM/SGLang); gRPC \`TokenizedInput\` stubbed behind the interface.
- \`PrefixCacheProbe\` canary: same token-ID prompt twice → warm prefix-cache hit proves the engine caches exactly those tokens. CI-verified vs a mock vLLM; gated live test + \`canary_dual_input_tokenization.sh\` for a real CPU engine.

**Build/CI**: default \`make ci\` stays pure-Go and green; the cgo tokenizer is opt-in (not in default CI). Logic coverage 90.7%.

**Not run here**: the live CPU-engine canary — the vLLM image pull fails under VPN TLS interception in this environment (documented caveat). The engine half is CI-verified against a mock; the tokenizer (real Qwen) and fingerprint (golden) halves are verified for real.

**Follow-ups** (non-blocking): expose engine block size as a flag (fixed 16 today); add an opt-in \`-tags smgcgo\` compile check to CI; multi-turn structured prompt_text input.


EdHasNoLife:
Addressed the round-1 review in 4391763:

- **Blocking** (inferencecache_service.go:163): the lookupTimeoutMs budget is now applied *before* prompt_text resolution, and dual-input resolution runs under that deadline (goroutine+select) → a slow first-time tokenizer load fails open with TIMEOUT. `resolveLookupChain` no longer mutates the request, so the bounded goroutine is race-free. New test `TestLookupRoutePromptTextSlowTokenizerTimesOut`.
- **Should-fix** (block size): added `--engine-block-size` (default 16) + `server.WithEngineBlockSize`; grpc-contract.md updated; test that a non-16 size changes chunking.
- **Should-fix** (canary.go:60): prefix_cache_queries scrape errors are now propagated, not discarded.
- **Nits**: dropped the unused `CompletionParams.Stream`; `PrefixCacheProbe.Run` guards a nil Client; `OpenAIClient.Complete` tolerates a nil http.Client.

Skipped one Nit (canary.go scrapeCounter absent-vs-zero): vLLM always exposes `vllm:prefix_cache_*_total` (starting at 0), so 'absent' isn't a real state here, and the canary's `HitsDelta <= 0` assertion already surfaces a missing/zero counter as a clear failure. Happy to add an explicit missing-metric error if you'd prefer.

EdHasNoLife:
Round-2 addressed in 903ec21:

- **Should-fix** (sub-block fallthrough, inferencecache_service.go:308/318): a token_ids / prompt_text input shorter than one full KV block now fails open with NO_HINT — no fall-through to the legacy empty-prefix / TENANT_HOT path. Test `TestLookupRouteShortTokenIDsFailsOpenNotTenantHot`.
- **Should-fix** (placement): moved `pkg/engineclient` → `pkg/adapters/engineclient`; package doc notes it's harness/egress code owned by no binary (peer to the ingest adapter under pkg/adapters/engine).
- **Should-fix** (cgo concurrency): documented that it's safe — upstream `Encoder`/`Decoder` are `Send + Sync` with `&self` methods, so concurrent Encode on the shared `Arc<dyn Tokenizer>` handle is fine; the cache lock only guards creation. Adding a per-handle mutex would needlessly serialize the hot path.
- **Nit** (scrapeCounter): now skips the {labels} block (label values may contain spaces) and reads the value token, ignoring a trailing timestamp. Test `TestScrapeCounterReadsValueNotTimestamp`.

EdHasNoLife:
Round-3 addressed in 021368a:

- **Should-fix** (token echo dropped, :206): the canonical token_ids are now echoed on **every** response after the server tokenized prompt_text — the min-prefix-gate NO_HINT and the post-tokenization TIMEOUT paths included — so a tokenizer-less gateway always gets the tokens to forward to the engine. Test `TestLookupRoutePromptTextEchoesTokensOnMinPrefixGate`.
- **Should-fix** (ctx error → NO_HINT, :322): after resolution the handler checks `ctx.Err()` and returns fail-open **TIMEOUT** (not NO_HINT) when the lookup budget expired, so a deadline-aware tokenizer's `context.DeadlineExceeded` maps to the right code.
- **Should-fix** (placement): documented the engine egress client in CONTRIBUTING.md's layout table (`pkg/adapters/engineclient`, no binary owner) so the adapter layout is explicit and not a silent competing root.
- **Nit** (cgo test comment): `make tokenize-cgo-test` now defaults `IC_TEST_TOKENIZER` (`TOKENIZE_CGO_TEST_MODEL`, override-able) so the real-tokenizer tests actually run — matching the test's comment.

Skipping one **Nit** (canary image `latest-*` tags): this matches the repo's existing CPU canaries (`canary_e2e.sh`, `canary_c2_reconcile.sh`, etc.) and VERSIONS.md, which use `latest-{arch}` for the local CPU dev image and pin a digest only for the GPU reference deploy. The `IMAGE` env var is already the override for pinning. Pinning only this one canary would make it inconsistent with the others; happy to pin all of them in a separate sweep if you'd prefer.

EdHasNoLife:
Round-4 addressed in the latest commit:
- **Should-fix** (sub-block echo, :337): a sub-block prompt_text now fails open with NO_HINT **and** echoes the canonical token_ids it tokenized, so a tokenizer-less gateway still gets the engine input. Test `TestLookupRoutePromptTextSubBlockStillEchoes`.
- **Nit** (orphaned comment, tokenize.go:62): removed; `New` is documented at its per-build-tag declarations.

EdHasNoLife:
Round-5 addressed in 8b12ad9:
- **Should-fix** (FFI panic boundary, lib.rs:130): the create / encode_text / encode_chat C-ABI bodies are now wrapped in `catch_unwind`; a tokenizer panic returns `IC_ERR_PANIC` instead of unwinding across cgo, and the Go side maps it to `ErrUnavailable` → fail-open NO_HINT (no server abort).
- **Should-fix** (temperature omitempty, openai.go:38): dropped `omitempty` so an explicit `temperature: 0` (greedy/deterministic, what the canary uses) reaches the engine. The OpenAI client test now asserts the explicit 0 on the wire.

EdHasNoLife:
Round-6 addressed in 6c7acf8:
- **Should-fix** (no-deadline tokenization blocking, :192): prompt_text tokenization is now bounded by a default timeout (`DefaultTokenizeTimeout`, overridable) when neither a caller deadline nor `lookupTimeoutMs` applies → slow first-time loads fail open with TIMEOUT. Test `TestLookupRoutePromptTextDefaultTokenizeTimeout`.
- **Should-fix** + **Nit** (latency=0, :190/:205): added a handler-level clock + `noHintResponse` helper so every post-tokenization fail-open / TIMEOUT response reports the real elapsed hot-path time on `lookup_route_latency_seconds`.
- **Should-fix** (prompt_text semantics, proto:82): the proto + grpc-contract.md now specify the exact shape — single `user` chat message + generation prompt; multi-turn / pre-rendered input is unsupported, with `token_ids` as the escape hatch.
- **Should-fix** (arbitrary loads, main.go:49): documented that an empty `--tokenizer-models-dir` resolves model_id as a path/HF id (on-demand download) and that production should set a vetted dir so an untrusted caller can't drive arbitrary tokenizer loads.

EdHasNoLife:
Round-7 addressed in 6e00faa — the cgo tokenizer is reworked to fail-closed + eager-load, which resolves the Blocking trio at the root:

- **Blocking** (path traversal / request-driven downloads / uncancellable hot-path loads): the tokenizer now **eager-loads at startup** from an explicit, vetted `--tokenizer-models-dir` (one subdir per model) and serves from an in-memory map. The lookup path never loads, downloads, or joins a request `model_id` onto a filesystem path — no traversal, no request-driven HF downloads, no hot-path loads, and the cache is bounded to the vetted set. Empty dir ⇒ tokenization disabled (Unavailable → NO_HINT); unknown model_id ⇒ fail open.
- **Should-fix** (token_ids CPU): capped at `MaxLookupTokens` (~1M) → oversized requests fail open.
- **Should-fix** (handleFor global lock / unbounded cache): gone — the map is read-only after startup; Encode holds the RLock across the cgo call so Close can't free mid-use; concurrent encodes share it.
- **Should-fix** (success latency): bounded-path success now reports total hot-path time (tokenization + lookup).
- **Should-fix** (contract side-effect-free): grpc-contract.md updated — eager-load means no per-request I/O/downloads, so the property holds; model_id confined to the pre-loaded set.

New hermetic tests (no network): fail-closed without a dir, unknown-model fail-open, oversized token_ids; real-tokenizer encode/chat-template stays env-gated.

Deferred to tracked follow-ups (per scope decision): the **smgcgo CI compile/link gate** and a **golden chat-template token test**.

EdHasNoLife:
Round-8 addressed in e206d78:
- **Should-fix** (namespaced model ids, cgo_smg.go:85): the loader now walks the tree and keys each model by its slash-separated path relative to the vetted dir, so engine-facing ids like `Qwen/Qwen2.5-0.5B-Instruct` resolve (was: top-level basename only). Confinement unchanged. Factored into `discoverModels`/`isModelDir` with a hermetic test.
- **Should-fix** (prompt_text cap, :363): the tokenized prompt_text result is now capped at `MaxLookupTokens` too → a huge raw prompt fails open instead of fingerprinting/echoing a giant slice.
- **Nits** (doc sync): the `--tokenizer-models-dir` flag help, `Config.ModelsDir` godoc, and the rust README all updated — empty dir disables server-side tokenization (fail open NO_HINT); loads are eager + confined.

EdHasNoLife:
Round-9 addressed in 029f63e:
- **Blocking** (one-sided chain, :346): `resolveLookupChain` now treats any chain field (block_hashes and/or block_token_counts) as an explicit attempt and requires matching parallel-array lengths — a one-sided/malformed chain fails open with NO_HINT instead of falling through to token_ids/prompt_text/TENANT_HOT. Test `TestLookupRouteCountsOnlyChainFailsOpen` (warm tenant → still NO_HINT).
- **Should-fix** (:191): the default tokenize timeout is installed only when prompt_text is the EFFECTIVE input (no higher-precedence chain/prefix_hash/token_ids).
- **Should-fix** (test): added the handler counts-only case.
- **Nit** (canary.go:127): label-set skipping is now quote-aware. Test `TestScrapeCounterQuoteAwareLabels`.
- **Nit** (:24): `DefaultTokenizeTimeout` comment updated (eager startup loads; bound is defense-in-depth on the in-memory encode).

EdHasNoLife:
Round-10 addressed in 80d5ecc:
- **Should-fix** (proto↔TLS-doc sync): `grpc-tls.md` updated — dual-input `prompt_text` means `LookupRoute` can carry raw prompt text on `:9090` (hashed transiently, never stored). The two stale "never prompt text" claims now reflect that this *raises* the port's data-sensitivity and *strengthens* the TLS/mTLS case.
- **Should-fix** (uncancellable cgo encode): raw `prompt_text` is now capped at `MaxPromptTextBytes` (~1 MiB) **before** entering the tokenizer, bounding worst-case in-flight encode work for a timed-out request. Test `TestLookupRouteOversizedPromptTextFailsOpen`.
- **Nit** (alias comment): corrected — the explicit-chain path does alias req's read-only slices; safe because `resolveLookupChain` only reads req and a post-timeout result is discarded.

EdHasNoLife:
Round-11 addressed in 8fb7212:
- **Should-fix** (smoke gate, proto:97): the default-install smoke now asserts BOTH dual-input LookupRoute paths at the deployment level — `token_ids` → NO_HINT (unknown model), and `prompt_text` → NO_HINT on the default tokenizer-less build (fail-open, never an error).
- **Should-fix** (smgcgo CI gate, cgo_smg.go:1): deferred per the agreed scope — tracked as a dedicated follow-up ticket (a CI compile/link job needs the Rust toolchain + archive, which is infra beyond this PR). The opt-in `make tokenize-cgo-test` covers it locally.
- **Nit** (service:219): dropped the stale "load/download on first use" comment — tokenizers are pre-loaded at startup; the goroutine bound exists because the cgo encode is uncancellable, not because of loads.
- **Nit** (chain.go:5): fixed the off-by-one — blockHashes[i] covers tokens 0..(i+1)*blockSize.

EdHasNoLife:
Round-12 addressed in ca58aa5:
- **Should-fix** (openai.go:103): a 2xx with zero choices is now a wrapped error, not a silent empty Completion. Test `TestOpenAIErrorsOnNoChoices`.
- **Should-fix** (canary.go:144): `scrapeCounter` now reports whether the metric was present; `requireCounter` errors on an absent vLLM counter so the probe can't read a renamed/disabled metric as a real zero. Test `TestScrapeCounterAbsentMetric`.
- **Nit** (Makefile:235): `make tokenize-cgo-test` now also `go build -tags smgcgo ./...` (compiles + links the tokenizer-enabled server), matching the implied scope.
- **Nit** (cgo_smg.go:201): `EncodeText` rejects an embedded NUL (C.CString would truncate); the JSON-based Encode path is unaffected.
- **Nit** (canary_test.go:261): the live canary cycles token ids within a conservative vocab window (`IC_ENGINE_VOCAB`) so small-vocab model overrides stay valid.

EdHasNoLife:
Round-14 addressed in 1d4485c (rebased onto main; full CI + install-smoke now run on this PR):
- **Should-fix** (goroutine/CPU pileup, :228): concurrent prompt_text tokenizations are now capped by a non-blocking semaphore (`MaxConcurrentTokenize`). A saturated/wedged tokenizer sheds load and fails open to NO_HINT instead of accumulating in-flight cgo work. Test `TestLookupRoutePromptTextShedsLoadWhenSaturated`.
- **Should-fix** (silent ignore, new_default.go:8): added `tokenize.Enabled()`; cmd/server now warns loudly when `--tokenizer-models-dir` is set on a non-smgcgo build (a wrong-image deployment no longer looks configured while silently failing open).
- **Should-fix** (canary metric drift, canary.go:234): the vLLM hits/queries metric names are configurable on `PrefixCacheProbe` (`IC_ENGINE_HITS_METRIC` / `IC_ENGINE_QUERIES_METRIC` in the live test) so a renamed metric does not break the canary.
- **Nits** (fail-closed wording): reworded to "closed to request-controlled loading" in new_cgo.go, the rust README, and the cgo test, so it does not read as contradicting the fail-open (NO_HINT) lookup contract.

EdHasNoLife:
Round-15 addressed in b732889:
- **Should-fix** (chat-template tools, lib.rs:181): the shim now renders with `tools`/`documents` **undefined** (None), not `Some(empty)` — many HF/Jinja templates branch on `tools is defined`, so an empty list could render differently than the engine does for a no-tools chat request and desync our tokens. Re-verified against the real Qwen tokenizer.
- **Should-fix** (handler→tokenizer contract, route_dual_input_test.go:25): new `recordingTokenizer` + `TestLookupRoutePromptTextPassesSingleUserMessageToTokenizer` assert the handler calls `Encode` with the request model, the prompt as a single `user` message, and `AddGenerationPrompt=true`.
- **Should-fix** (golden token test, cgo_smg_test.go:40): deferred per the agreed scope — tracked as a dedicated follow-up (a model-version-pinned golden, ideally with a committed offline fixture, fits the smgcgo CI gate follow-up).
- **Nit** (fail-closed wording): reworded the last occurrence to match the fail-open contract.

EdHasNoLife:
Round-16 addressed (two Nits, no Blocking/Should-fix):
- **Nit** (cgo_smg.go:67): fixed the stale "each immediate subdirectory" comment — a model is any directory at any depth holding tokenizer artifacts, keyed by its relative path (so namespaced ids nest).
- **Nit** (openai_test.go:97): `TestGRPCClientNotImplemented` now asserts `errors.Is(err, ErrNotImplemented)`, pinning the stub sentinel.

heymrbox (approved):

PR #109, opened 2026-06-10
test(e2e): per-PR GPU-free gate for content-fingerprint routing

EdHasNoLife:
## What

A per-PR, GPU-free end-to-end regression gate for the content-fingerprint routing path, locking in the all-NO_HINT fix from #105 (this PR is stacked on that branch):

- **`cmd/kvevent-fake-engine`** — a test-only ZMQ PUB stand-in for a vLLM engine. Publishes a deterministic sequence of msgpack `BlockStored` events (token sequence chunked into blocks, optionally split across parent-chained events) on a configurable endpoint, republishing to cover the PUB/SUB slow-joiner race, and logs the expected content fingerprints (`EXPECT prefix_hash[...]`) for smoke assertions. An `-omit-token-ids` flag reproduces the engine-regression shape.
- **`e2e_test.go`** — runs the real pipeline end to end: fake engine over a **real ZMQ socket** → the **real kvevent-subscriber components** (`engine.Subscriber` → `engine.Reporter`, fingerprint derivation in `positionalIndex`) → the **real server** (`pkg/server`) over loopback TCP gRPC → `LookupRoute`.
- **CI**: dedicated per-PR job `fingerprint-e2e` (→ `make test-fingerprint-e2e`), added to the `Lint, Test, Proto` aggregate and the image-build gate. Unlike the scheduled kind canaries (which pull real engine images), this needs no GPU, no image, no cluster — it finishes in seconds, so it can gate every PR.

## Asserts (gateway's perspective, fingerprints recomputed independently via `pkg/fingerprint`)

1. `PREFIX_MATCH` with the publishing replica as the hint and the **full two-block chain** matched — including a block delivered in a second, parent-chained event (cross-event chaining over the wire).
2. A shorter one-block prefix of the same prompt also hits.
3. A **novel** prompt's fingerprint → `NO_HINT`, zero replica scores.
4. The engine's **own block hashes are not index keys** — the exact shape of the original bug.
5. **Producer contract**: `BlockStored` without `token_ids` logs `BlockStored produced no index entries` and indexes nothing (`GetCacheState` total = 0, would-be fingerprint → `NO_HINT`) — catches a vLLM version that stops emitting `token_ids` before it silently regresses routing to all-NO_HINT.

## Verification

- `make test-fingerprint-e2e` green; full `make ci` + `make pre-pr` + `make cover-check` (91.2%) green; e2e run 5× under `-race` with no flakes.
- Mutation-checked: breaking the query side (wrong `hash_scheme`) fails the test with a timeout, so the gate is not vacuously green.
- Unit tests round-trip every payload shape through the real `engine.DecodeEventBatch` (single-event, chained multi-event, omit-tokens), so the synthetic encoding can't drift from the decoder.


EdHasNoLife:
Re the round-6 nit (e2e also runs under `go test ./...` / `make test-race`): intentional, keeping it. The duplication costs ~2s; in exchange every plain `go test ./...` run — local or CI — carries the regression lock with no opt-in to forget, and the test code itself can never silently bit-rot behind a build tag. The dedicated `fingerprint-e2e` job exists for visibility (a named per-PR gate with verbose output), not exclusivity.

heymrbox (approved):

EdHasNoLife:
Re the round-9 nits, both deliberate:

- **E2E also runs under `go test ./...`** (re-raised from earlier rounds): keeping it — see the rationale in https://github.com/cachebox-project/inference-cache/pull/109#issuecomment-4674608589. The duplication is ~2s and means the regression lock travels with every plain test run, with no opt-in to forget. (The 20s deadlines are worst-case timeouts, not run time — the tests converge in ~60ms.)
- **`test-fingerprint-e2e` filters to `-run 'TestE2E'`**: also deliberate. The target/job is the routing-path gate specifically; the payload↔decoder round-trip tests are unit tests that ride the ordinary `make test`/`make test-race` paths (same CI pipeline, same PR gate). Since the e2e itself round-trips every payload through the real decoder over the wire, decoder drift fails the gate anyway.

PR #110, opened 2026-06-10
Retire spec.storage.pvc — durability is a backend choice

EdHasNoLife:
Retire `spec.storage.pvc` + `status.capacity` from CacheBackend at v1alpha1. Durability is a **backend choice** (the planned Mooncake backend), not a generic volume knob — see `docs/design/lmcache-server-persistence.md` (PR #106) for the decision record and evidence.

## Why
- The `lm://` lmcache-server we provision is **in-memory**; a local PVC cannot back it, and the disk-capable MP-mode server's transfer path is shared-memory-bound (doesn't fit the per-backend Deployment + ClusterIP model).
- The PVC plumbing shipped in #82 provisioned + mounted a volume the server never wrote to — an honest-but-empty promise; carrying it to v1beta1 locks in a category error.
- Removal at pre-launch v1alpha1 is permitted by the schema-evolution policy (inert/declarative field, no external consumers).

## What changed (+140 / −1941)
- **API**: `spec.storage{,.pvc}`, `status.capacity`, `CacheBackendStorageSpec`/`CacheBackendPVCSpec` removed; `ResolveCacheServer` back to `(*PodSpec, *Service, error)` (drops `ResolvedCacheServer`/`AdapterDataVolume`).
- **Controller**: PVC provisioning/mount, multi-replica `InvalidStorageConfiguration` gate, adopt-and-keep + `OrphanedPVCRetained`, capacity reporting, `Owns(PVC)`, PVC RBAC all removed. Server-restart cascade's lifecycle-exit set is now External/Unmanaged.
- **Webhook**: memory-only / size>0 / name-length storage rules + `memoryOnlyBackends` removed; aggregation test rewritten on surviving rules.
- **Tests**: storage unit + envtest suites deleted; CRD contract test now guards `spec.storage` + `status.capacity` **absent**; harness/mocks reverted to the tuple.
- **Docs/samples/smoke**: api + crd-contract rows, `recipe-persistent-cache.yaml` + catalog entry, lmcache-sample disclaimer (now points at the decision doc), install-smoke storage phase.
- Regenerated CRD + RBAC + deepcopy.

## Merge order
Merge **#106 first** (decision doc) — this PR's docs reference `docs/design/lmcache-server-persistence.md`.

## Test plan
- `make pre-pr` green (race tests, drift-free regen, verify-samples 16 ok).
- CRD contract test pins the retired fields absent (guards re-introduction).
- Evolved server-restart cascade suite passes unchanged apart from the storage-exit case removal.


EdHasNoLife:
On the decision-doc dependency finding: deliberate and documented — this PR's body specifies the merge order (**#106 first**, which adds `docs/design/lmcache-server-persistence.md`, then this PR). The references here are correct once that lands; making this PR self-contained would duplicate the doc and conflict at merge. The other two findings are fixed in the latest commit.

EdHasNoLife:
Addressed the review (commit `75d5550`):

- **Should-fix (install-smoke):** added CRD-schema-trim assertions that the installed CRD no longer serves `spec.storage` or `status.capacity` — the operator-facing surface change now has a real default-install assertion, alongside the existing `lookupTimeoutMs`/`minimumPrefixTokens`/`indexEntries` absence checks.
- **Should-fix (decision-record dead link):** `docs/design/lmcache-server-persistence.md` is the deliverable of the companion decision PR #106; this PR depends on it and the **Merge order** section already calls out merging #106 first. Not duplicating the doc here (it belongs to #106) — the two are a co-dependent CAC-127/CAC-182 pair.
- **Nit (stale comment):** corrected the `cachebackend_types_test.go` comment — `status.health` was removed in an earlier change; this PR removes `status.capacity`.

EdHasNoLife:
Round 2:
- **Nit (smoke comment):** fixed in `976f56a` — the schema-trim comment now says "five removed fields" (the storage/capacity additions).
- **Decision-doc findings (cachebackend-api / sample):** `docs/design/lmcache-server-persistence.md` is the deliverable of companion PR #106 (CAC-127), now rebased onto current main and green. The **Merge order** section already enforces "merge #106 first"; not duplicating the doc here since it belongs to #106 — the two are a co-dependent CAC-127/CAC-182 pair.

heymrbox (approved):

PR #112, opened 2026-06-12
Add StatefulSet CacheBackend support

heymrbox:
## Summary

Adds StatefulSet support for managed CacheBackend workloads. `deploymentKind: StatefulSet` now reconciles an owned StatefulSet, targets it from HPA, reports readiness from ready replicas, and uses `volumeClaimTemplates` for per-replica PVCs when `spec.storage.pvc` is set.

Also updates generated RBAC/CRD docs, samples, design docs, controller/unit/envtest coverage, and install-smoke assertions for the StatefulSet persistent-storage path.

## Linked issues

CAC-115

### Quality
- [ ] CI is green.


heymrbox:
Addressed in a560a88.

- StatefulSet storage edits now surface explicit immutable drift: when the desired volumeClaimTemplates shape differs from the owned live StatefulSet, the controller leaves the StatefulSet untouched, clears the advertised endpoint, and publishes Ready=False / Progressing=False with reason ImmutableStatefulSetStorage.
- The Deployment restart-cascade polling cadence now stays on Deployment-managed backends only, so StatefulSet backends do not self-requeue every 30s for cascade work they cannot perform.
- Added direct tests for StatefulSet ReadyReplicas readiness, StatefulSet no-cascade requeue behavior, and immutable storage drift; updated the CacheBackend design doc with the new status reason.

Verification:
- go test -count=1 ./internal/controller -run "TestReconcileStatefulSet(StorageEditSurfacesImmutableDrift|DoesNotScheduleDeploymentCascadePoll)|TestManagedReadinessForStatefulSetUsesReadyReplicas"
- go test -count=1 ./internal/controller
- make pre-pr

heymrbox:
Addressed in f5fe047.

- Added a SharedPVCRetained warning when a persistent Deployment backend switches to StatefulSet and the prior shared Deployment PVC is retained but no longer mounted.
- Immutable StatefulSet storage drift now clears the probe rate-limit slot before removing FunctionalProbeOK, so returning to the managed path gets a fresh probe instead of reusing a stale limiter entry.
- Scoped the persistent-cache sample comment to Deployment-backed ReadWriteOnce PVCs and documented the SharedPVCRetained event in the CacheBackend design notes.

Verification:
- go test -count=1 ./internal/controller -run "TestReconcile(StorageKeptViaStatefulSetSwitchWarnsAndKeepsDeploymentPVC|ImmutableStatefulSetStorageClearsProbeRateLimiter)"
- go test -count=1 ./internal/controller
- make pre-pr

heymrbox:
Addressed in f9bfbc1.

- Normalized StatefulSet volumeClaimTemplate drift comparison so apiserver-defaulted volumeMode=Filesystem does not trip ImmutableStatefulSetStorage.
- Made per-replica StatefulSet PVC retention explicit as Retain/Retain and documented the retention/reclaim behavior.
- Extended install-smoke to force a follow-up reconcile and assert persistent-cache-sts status.endpoint plus Ready=True/BackendReady stay published.

Verification:
- go test -count=1 ./internal/controller -run "Test(ReconcileStatefulSetPersistentStorageUsesVolumeClaimTemplates|StatefulSetVolumeClaimTemplateComparisonIgnoresAPIServerDefaults)"
- go test -count=1 ./internal/controller
- bash -n docs/reference-stack/scripts/default_install_smoke.sh
- make pre-pr

heymrbox:
Addressed in ed1b9e3.

- StatefulSet updates now reconcile persistentVolumeClaimRetentionPolicy back to the documented Retain/Retain contract if it drifts or was created before the field was set.
- Split retained-PVC warning guards by reason, so OrphanedPVCRetained and SharedPVCRetained can both fire once for the same retained PVC when it moves through both operator-visible states.
- Updated the stale KV-event gate comment to describe managed workloads instead of only Deployments.

Verification:
- go test -count=1 ./internal/controller -run "TestReconcile(StatefulSetRepairsPVCRetentionPolicy|RetainedPVCWarningsAreReasonScoped)"
- go test -count=1 ./internal/controller
- make pre-pr

heymrbox:
Addressed in e6b4535.

- Kind-switch replacement failures now still run the managed status pass with an empty endpoint and non-Ready workload observation, so a deleted old workload is not advertised if the replacement apply fails.
- Added regression coverage for both Deployment→StatefulSet and StatefulSet→Deployment replacement failures.
- Updated the controller and CacheBackend API docs to say managed workloads can be Deployment or StatefulSet.

Verification:
- go test -count=1 ./internal/controller -run "TestReconcile(DeploymentToStatefulSetApplyFailureClearsStatus|StatefulSetToDeploymentApplyFailureClearsStatus)"
- go test -count=1 ./internal/controller
- make pre-pr

heymrbox:
Addressed in 2484b08.

- StatefulSet-managed backend pods now participate in the same server-instance cascade path as Deployment-managed backends. The observer authenticates StatefulSet pods through their direct ownerRef, tracks their `<pod-uid>:<restart-sum>` identity, and still rolls injected engine Deployments on a real cache-server replacement.
- Kept the periodic server-instance poll for StatefulSet backends so in-place cache-server container restarts are observable there too.
- Updated the API/design docs and regenerated the CacheBackend CRD descriptions for `capacity`, `firstAvailableAt`, and `observedServerInstance`.
- Added regression coverage for a StatefulSet cache-server pod replacement cascading the injected engine Deployment, and updated the old StatefulSet skip/clear expectations.

Verification:
- go test -count=1 ./internal/controller -run "TestReconcileServerInstance_StatefulSetPodReplacementCascadesEngineDeployment|TestReconcileStatefulSetSchedulesServerInstancePoll|TestReconcileSwitchToStatefulSetShedsDeployment"
- go test -count=1 ./internal/controller
- make pre-pr

heymrbox:
Addressed in 49c82e5.

- Updated the CacheBackend API spec table so `storage.pvc.size` and `storage.pvc.storageClassName` describe the two managed storage paths separately.
- Deployment-backed shared PVCs still document the upward-size patch / immutable StorageClass behavior.
- StatefulSet-backed per-replica PVCs now explicitly document that size or StorageClass changes alter `volumeClaimTemplates` and surface as `Ready=False/ImmutableStatefulSetStorage` instead of being applied in place.

Verification:
- make pre-pr

heymrbox:
Addressed in 9564625.

- Immutable StatefulSet storage drift now still reconciles the owned HPA, so clearing or changing `spec.autoscaling` is not skipped while the controller is publishing `Ready=False/ImmutableStatefulSetStorage`.
- The drift path no longer clears the server-instance latch/shadow for a live StatefulSet. It keeps the normal server-instance observer and periodic poll active, so cache-server restarts during drift can still cascade to injected engine Deployments.
- Added regressions for both cases: HPA cleanup during drift, and server restart cascade during drift.

Verification:
- go test -count=1 ./internal/controller -run "TestReconcile(StatefulSetImmutableStorageDriftStillCleansUpHPA|ServerInstance_ImmutableStatefulSetStorageDriftStillCascades)"
- go test -count=1 ./internal/controller
- make pre-pr

heymrbox:
Addressed in 88509f4.

- Added adapter-aware admission validation for persistent StatefulSet backends: the validator now checks the actual generated per-replica PVC name shape, `<volumeClaimTemplate>-<statefulset>-<ordinal>`, against the Kubernetes 253-character object-name limit.
- The check uses the selected runtime adapter’s declared `DataVolume` name instead of hard-coding the current LMCache `cache-data` template name.
- Added a regression where the CacheBackend name passes the Deployment shared-PVC check but would make the StatefulSet PVC name too long.

Verification:
- go test -count=1 ./internal/webhook/v1alpha1 -run TestValidator_StorageWithOverlongNameRejected
- go test -count=1 ./internal/webhook/v1alpha1
- make pre-pr

heymrbox:
Addressed in b0e082a.

- Added the StatefulSet generated-PVC-name validation to the CacheBackend design doc’s validating rules table.
- The row documents both child PVC name shapes: Deployment shared PVCs as `<cachebackend-name>-data`, and StatefulSet per-replica PVCs as `<volumeClaimTemplate>-<statefulset-name>-<ordinal>` using the selected runtime adapter’s `DataVolume` name.

Verification:
- make pre-pr

PR #116, opened 2026-06-16
Audience-bind policy endpoint tokens

heymrbox:
## Summary

Audience-binds the policy push path separately from the controller snapshot/probe path. The server now accepts `/policy` writes only with the policy audience token, while `/snapshot` and `/probe` continue to use the controller audience. The default install mounts both projected service account tokens, wires the matching server flags, updates the policy pusher default token path, and extends the smoke probe plus docs for the split-audience auth model.

## Linked issues

CAC-113

### Quality
- [x] CI is green.


pkg/server/server.go
```diff
@@ -257,18 +269,20 @@ func New(opts ...Option) *Service {
...
 	// caches are independent — one extra TokenReview per endpoint per TTL
-	// window in the steady state, negligible vs the apiserver's own auth
-	// cost. The shared Audience means the projected SA token at the
-	// controller's volume mount admits to all three endpoints uniformly;
-	// there's one trust boundary, not three.
+	// window in the steady state, negligible vs the apiserver's own auth cost.
 	if s.controllerAuthCfg != nil {
+		policyAudience := s.policyAudience
+		if policyAudience == "" {
+			policyAudience = s.controllerAuthCfg.audience
```
EdHasNoLife, line 284: Note (not blocking): the empty-string fallback means a future caller that wires `WithControllerAuth` but forgets `WithPolicyAudience` would silently re-couple `/policy` to the controller audience — the exact coupling this PR removes — with no error or warning. The production binary always pairs the two (`cmd/server/main.go`) and the godoc documents this as intentional backwards-compat for in-process tests, so it's fine today. Just calling it out as a latent footgun; a one-line WARN log when the fallback path is taken under a non-test config would make a future misconfiguration self-announcing.

EdHasNoLife (commented):
Reviewed the full diff against the project's design context (PROJECT_CONTEXT.md, `policy-propagation.md`, the `pr-review` checklist) and traced the audience split end-to-end. **No blocking or should-fix findings** — this is a careful, well-tested change. Two low-priority advisory notes below.

### What I verified myself (isolated `pr-116` worktree)
- `go build ./...` — pass
- `go vet ./pkg/server/... ./internal/controller/... ./cmd/...` — pass
- Unit tests on all changed packages (`pkg/server`, `pkg/server/auth`, `internal/controller`, `cmd/server`) — pass
- **Envtest integration against a real apiserver** — pass, including the cross-endpoint check that a controller-audience token is rejected on `/policy` (`cachepolicy_authed_integration_test.go`) and the apiserver-backed audience enforcement in `pkg/server/auth`

### End-to-end consistency (by inspection)
- Production wiring is correct: the policy pusher leaves `BearerTokenPath` empty → `DefaultPolicyBearerTokenPath` (policy-token); the snapshot poller and `ProbeClient` → `DefaultBearerTokenPath` (controller-token). Server authenticators: snapshot → controller, **probe → controller**, policy → policyAudience. Each endpoint's enforced audience matches the token audience projected in `config/manager/manager.yaml`.
- `manager.yaml` renders as valid YAML with both projected token volumes intact (`audience` + `expirationSeconds` on each).
- Smoke body `{"version":5,...,"tenants":[]}` matches the current `PolicyPropagationVersion` (min accepted 3), so the `204` happy-path assertion is valid.

### Note 1 — upgrade-ordering transient (not a fix; flagging for awareness)
Because the audiences are now distinct, a rolling upgrade has a window where policy pushes 401 until **both** the controller and the server are on the new version (a new controller sends a policy-audience token to an old server that still expects the controller audience, and vice-versa). `/snapshot` + `/probe` are unaffected. It self-heals via soft-state + the 30s re-push, and pre-launch there's no zero-downtime guarantee — so this is acceptable as-is; just worth being aware of (no migration shim accepts both audiences on `/policy`).

### Note 2 — see inline comment on `pkg/server/server.go`
The `WithPolicyAudience` empty-string fallback is a (documented, intentional) latent footgun.

EdHasNoLife (approved):

PR #119, opened 2026-06-17
feat: deploy-time LMCache kernel-health gate (EngineKernelsHealthy)

EdHasNoLife:
Detects the lmcache native `c_ops` CUDA-kernel load failure — the verified root cause of silent T2 reload serialization — at engine-pod **deploy time**, and surfaces it on `CacheBackend.status`.

**What**
- Injected `lmcache-kernel-check` init container force-loads the native `c_ops*.so` in the engine's own image. A naive `import lmcache.c_ops` is **stubbed** by lmcache's fallback shim → always succeeds (proven from source); force-loading reproduces `libcudart.so.13` empirically, GPU-free.
- New `EngineKernelsHealthy` condition. **Default fail-open** (report-only; does NOT gate `Ready`). **Opt-in strict** (annotation `inferencecache.io/lmcache-kernel-check: strict`) hard-fails the pod in `Init` + downgrades `Ready` (reason `EngineKernelDegraded`).
- `auto` mode injects only on GPU engine pods (CUDA-only). False reason `KernelLoadFailed` (specific cause in `.message`); `KernelCheckError` / `KernelCheckPending` buckets; CrashLoopBackOff handled.

**Boundary**: green = kernels *loaded*, not *fast under concurrency* — load-time linkage, not runtime executability.

**Scope**: local/c_ops mode only. **No CRD/proto schema change** (`EngineKernelsHealthy` is a generic condition). Follow-ups: CAC-196 (GPU concurrency canary — the required symptom validator) and CAC-197 (server-side `lm://` prober).

**Tests**: adapter + webhook unit; controller unit + envtest (report-only no-downgrade / strict downgrade); hermetic detector-script test; GPU-free install-smoke (injection shape + FAIL→condition).

Closes CAC-194.

Naming: vendor-neutral, no oci/oracle in core identity. Proto/CRD unchanged → no v1alpha1 back-compat concern.


EdHasNoLife:
Addressed the Codex review (commit `8af49a2`):

- **Blocking (detector false-positive on healthy engines):** switched the detector from `importlib.exec_module` to `ctypes.CDLL`. `exec_module` derives `PyInit_<spec-name>`, so loading the real `c_ops.so` under any other module name would `FAIL` a *healthy* engine (only the libcudart-missing path masked it, since that fails at dlopen first). `ctypes.CDLL` forces the same dlopen/DT_NEEDED resolution — where the CUDA mismatch lives — with no init-symbol coupling. Re-verified against the real PyPI wheel: still raises `libcudart.so.13` with torch present. Added `TestKernelCheckScriptHealthyExtensionReportsOK` (compiles a trivial `c_ops.so`) as the OK-path regression guard the old loader would have failed.
- **Should-fix (image pull policy):** the init container now copies the engine container's `ImagePullPolicy` instead of hard-coding `IfNotPresent`.
- **Should-fix (PSA admissibility):** copies the engine container's `SecurityContext` so appending the init container can't make a restricted-PSA engine pod fail admission (preserving fail-open).
- **Nit (hermeticity):** detector-script test now runs `python -S`.

Also fixed the install-smoke failure from the prior push (the report-only stand-in pod now survives the webhook-injected `--kv-transfer-config` arg via `sh -c`).

EdHasNoLife:
Round 5 addressed (commit `7b05a6e`):
- **#2 (annotation-flip):** the strict `Ready` downgrade is now derived from each FAILing pod's *admitted* mode (its kernel-check init container's `KERNEL_CHECK_STRICT` env), not the CacheBackend's current annotation. Flipping the annotation on a live backend no longer misreports pods that are actually running report-only, and self-corrects as pods roll. Pod-truth unit + integration tests added.
- **#3 (invalid annotation):** rejected at admission now (new CacheBackend validation rule + `IsValidKernelCheckMode`), so a typo can't silently relax strict → report-only. Unit test + real-install smoke assertion added.
- **#1 (readOnlyRootFilesystem masking) — empirically disproven, no change.** `/dev/termination-log` IS writable under a read-only root filesystem (`/dev` is a separate mount, unaffected by `readOnlyRootFilesystem: true`; verified with `docker run --read-only`). The detector's report channel works fine with the engine's SecurityContext copied — no masking occurs.

heymrbox (approved):

PR #121, opened 2026-06-19
Add CacheBackend binding diagnostics

heymrbox:
## Summary

Adds operator-facing diagnostics for CacheBackend to engine-pod binding issues:
- surfaces unmatched `engineSelector` drift through `status.engineSelectorMessage` and `EngineSelectorUnmatched` Events
- stamps explicit skip-inject pods with `inferencecache.io/inject-skipped` and emits `SkippedByOperator` Events
- uses a faster matched-pod refresh cadence during observed-vs-desired Deployment churn without adding a Pod watch
- extends docs, envtest/unit coverage, generated CRD schema, and install-smoke assertions for the new surfaces
## Linked issues

CAC-91

### Quality
- [x] CI is green.


heymrbox:
Addressed in b0df6a4.

I moved the engine-pod match cadence rationale out of the unlinked root note and into `docs/design/cachebackend-api.md`, then removed `PROJECT_CONTEXT.md`.

I also updated the design contract docs so they now cover:
- `status.engineSelectorMessage` in the CacheBackend status table
- CacheBackend status ownership in `docs/design/crd-contract.md`
- `inferencecache.io/inject-skipped` stamping and clearing behavior
- the `SkippedByOperator` pod Event and `EngineSelectorUnmatched` CacheBackend Event

Verification:
- `make pre-pr`

heymrbox:
Addressed in c6a9251.

I tightened `SkippedByOperator` so the controller now requires both pieces of evidence before emitting the event: a truthy `inferencecache.io/skip-inject` opt-out and the webhook-stamped `inferencecache.io/inject-skipped: skip-inject-annotation` marker. A stale or pre-set `inject-skipped` marker by itself now gets ignored and is filtered by the pod-event predicate.

I also updated the fail-open helper godoc to name both decision annotations, and aligned the CacheBackend design doc with the stricter skipped-event contract.

Verification:
- focused controller regression test first failed on the forged-marker case, then passed after the fix
- `go test ./internal/controller -run 'TestEnginePodEvents_(NoSkippedEventOnForgedInjectSkippedMarker|PredicateIncludesInjectedAndSkippedPods|EmitsSkippedEventOnSkippedPod)'`
- `go test ./internal/webhook/pod -run 'TestSkipAnnotationOptsOut|TestHandle_FailOpenClearsStaleInjectSkipped|TestHandle_SkipAnnotation'`
- `make pre-pr`

heymrbox:
Addressed in 8c829c2.

I corrected the CacheBackend design doc so stale annotation stripping is described only for fail-open paths after the webhook has decoded a Pod. The doc now calls out that decode failures fail open before there is a Pod object to patch, so stale annotations cannot be cleared on that path.

I also renamed `TestHandle_SkipAnnotation_Passthrough` to `TestHandle_SkipAnnotation_StampsInjectSkipped`, which matches the current contract: skip-inject now stamps `inferencecache.io/inject-skipped`.

Verification:
- `go test ./internal/webhook/pod -run 'TestHandle_SkipAnnotation_StampsInjectSkipped'`
- `make pre-pr`

heymrbox:
Addressed in 99dcc51.

I changed the matched-pod churn detector so desired replica comparison is only used when the controller actually finds at least one matching Deployment template. Matching bare pods, StatefulSet pods, Jobs, or other non-Deployment owners no longer force the CacheBackend onto the 5s churn cadence forever.

I added a regression test for matching pods with no matching Deployment: it failed on the previous code with a 5s requeue and now passes with the steady 30s cadence. The existing Deployment desired-replica mismatch test still covers the fast-cadence path.

Verification:
- red/green focused cadence test
- `go test ./internal/controller -run 'TestReconcileMatchedEnginePodsUses(ChurnCadenceWhenDeploymentDesiredDisagrees|SteadyCadenceWithoutMatchingDeployment)'`
- `go test ./internal/controller`
- `make pre-pr`

heymrbox:
Addressed in f8a75f5.

I gated the unmatched selector diagnostic on expected desired replicas. When a matching Deployment is intentionally scaled to zero, `status.matchedEnginePods` still reports the observed `0`, but `status.engineSelectorMessage` stays empty and the controller does not emit `EngineSelectorUnmatched`.

The existing no-owner zero-match diagnostic is preserved, and a matching Deployment with desired replicas > 0 and zero pods still gets the unmatched message/Event.

Verification:
- red/green scaled-to-zero diagnostic test
- `go test ./internal/controller -run 'TestReconcileMatchedEnginePods(NoUnmatchedDiagnosticWhenDeploymentScaledToZero|ZeroWhenSelectorMatchesNothing)'`
- `go test ./internal/controller`
- `make pre-pr`

heymrbox:
Addressed in 724fa57.

I aligned the status contract with the scaled-to-zero exception: `engineSelectorMessage` now documents that a matching Deployment scaled to zero still reports `matchedEnginePods: 0`, but does not surface the drift message because no engine pods are expected. The generated CRD description was regenerated to match.

I also updated the event guard so `EngineSelectorUnmatched` fires when an older object already has `status.matchedEnginePods: 0` but gains the new diagnostic message during upgrade. Steady-state zero with an unchanged message still does not re-emit.

Verification:
- red/green upgrade-event regression test
- `go test ./internal/controller -run 'TestReconcileMatchedEnginePods(EmitsUnmatchedEventWhenMessageAddedToExistingZero|EmitsUnmatchedEventOnInitialZeroAndTransition|NoUnmatchedDiagnosticWhenDeploymentScaledToZero|ZeroWhenSelectorMatchesNothing)'`
- `go test ./internal/controller`
- `make pre-pr`

heymrbox:
Addressed in e328638.

I split the desired-replica lookup result so the reconciler can distinguish “no matching Deployment exists” from “Deployment List failed.” On a Deployment-list error, the controller now still records the observed `matchedEnginePods` count, but preserves the existing `engineSelectorMessage` and suppresses `EngineSelectorUnmatched` rather than creating a fresh false drift diagnostic.

I also added `status.engineSelectorMessage` to the CacheBackend CRD schema sentinel and deepcopy coverage.

Verification:
- red/green Deployment-list-error diagnostic regression test
- `go test ./internal/controller -run 'TestReconcileMatchedEnginePodsNoUnmatchedDiagnosticOnDeploymentListError'`
- `go test ./internal/controller -run 'TestReconcileMatchedEnginePods(NoUnmatchedDiagnosticOnDeploymentListError|EmitsUnmatchedEventWhenMessageAddedToExistingZero|NoUnmatchedDiagnosticWhenDeploymentScaledToZero|ZeroWhenSelectorMatchesNothing)'`
- `go test ./api/v1alpha1 -run 'TestCacheBackendCRDSchemaFieldsAndEnums|TestCacheBackendDeepCopyCopiesNestedFields'`
- `make pre-pr`

heymrbox:
Addressed in ad7ab1a.

I updated the served `matchedEnginePods` godoc so per-pod visibility points to both `InjectedByCacheBackend` for injected pods and `SkippedByOperator` / `inferencecache.io/inject-skipped` for explicit opt-outs. I regenerated the CacheBackend CRD description to match.

I also clarified the design doc: `SkippedByOperator` means the persisted pod carries the truthy opt-out plus skipped marker, not cryptographic proof that the webhook authored the marker.

Verification:
- `make manifests`
- `make pre-pr`

internal/controller/cachebackend_controller.go
```diff
@@ -2053,20 +2084,96 @@ func (r *CacheBackendReconciler) refreshMatchedEnginePods(ctx context.Context, b
...
 				"namespace", backend.Namespace, "name", backend.Name, "error", err.Error())
-			return
+			return out
 		}
 		count := int32(len(pods.Items))
-		if backend.Status.MatchedEnginePods != nil && *backend.Status.MatchedEnginePods == count {
-			return
+		desired, desiredKnown, desiredReliable := r.desiredEngineReplicas(ctx, backend, matcher)
+		selectorDiagnosticReliable = desiredReliable
+		if desiredReliable && desiredKnown && desired != count {
```
EdHasNoLife: **Should-fix — churn cadence can stick at 5s forever (defeats the documented "steady state stays 30s").**

`count` (line 2089) is `len(pods.Items)` from a label-only List with **no phase filter**, so it counts pods in *every* phase — including `Failed`/terminal pods. `desired` (line 2090) is the live Deployment replica sum. The churn flag is `desired != count`.

Kubernetes does not promptly GC terminal pods: `--terminated-pod-gc-threshold` defaults to 12500, so an `Evicted`/`Failed` pod that still carries the engine labels lingers essentially forever in a small cluster. vLLM engine pods are memory-hungry and node-pressure eviction (phase `Failed`, reason `Evicted`) is a routine production event — a single lingering Evicted pod makes `count` (e.g. 4) permanently exceed `desired` (3).

**Failure scenario:** a vLLM engine Deployment (replicas=3) has one pod evicted under node memory pressure → the Evicted pod keeps matching the selector → `desired(3) != count(4)` stays true every reconcile → the CacheBackend requeues at the 5s churn cadence indefinitely (≈6× the intended reconcile rate, each doing an uncached Pod List + Deployment List), and never reverts to 30s. `TestReconcileMatchedEnginePodsUsesChurnCadenceWhenDeploymentDesiredDisagrees` only exercises live pods, so this path isn't covered.

Consider comparing `desired` against a non-terminal pod count (filter out `Succeeded`/`Failed`, or count only `Running`+`Pending`) for the churn decision, so terminal-phase residue doesn't pin the fast cadence on. The operator-facing `matchedEnginePods` snapshot can stay all-phase if that's the intended CAC-87 semantics; it's only the churn comparison that needs the terminal-phase exclusion.

EdHasNoLife (commented):
## Review — CAC-91 CacheBackend binding diagnostics

Reviewed the full diff against `PROJECT_CONTEXT.md` (C-series substrate, §5), the `pr-review` checklist, and the CAC-91 Definition of Done. Solid, well-scoped PR that delivers all three CAC-91 surfaces.

**Verified locally** (worktree on `ad7ab1a`):
- `go build ./...`, `go vet ./internal/...` — clean
- `go test ./internal/controller/... ./internal/webhook/...` — all pass
- `make manifests generate` — **no drift** (committed CRD matches generated)
- RBAC for the new `Deployment` List is already present in `config/rbac/role.yaml` (apps/deployments get;list;watch)

**DoD coverage (checked by inspection):**
- ✅ `status.engineSelectorMessage` echoes the selector when matched=0 with a configured selector; cleared on match / scaled-to-zero / selector removal.
- ✅ `EngineSelectorUnmatched` Normal Event on initial-zero, 1+→0, and the upgrade-from-older-zero-status case; suppressed on steady state (early-return when count+message unchanged).
- ✅ `inferencecache.io/inject-skipped: skip-inject-annotation` stamped by the webhook skip path; `SkippedByOperator` Event gated on *both* a truthy `skip-inject` and the webhook's marker.
- ✅ Conditional 5s churn cadence (no Pod watch added — Pod List stays uncached via APIReader; Deployment desired-replica List added).
- ✅ Concept-doc failure-mode table updated (skip vs drift distinction); `cachebackend-api.md` / `crd-contract.md` status surfaces updated.
- ✅ install-smoke extended with both the unmatched-selector and skip-inject cases.

**Security note (not a finding):** the forged-annotation surface is handled consistently with the existing `injected-by` posture — `skipInjection` strips a forged `injected-by`/`injected-by-uid` pair, `failOpen` now also strips `inject-skipped`, and `SkippedByOperator` requires both annotations. The `failurePolicy=Ignore` window (forged annotations admitted while the webhook is down) is pre-existing and equally bounded for the new skip path; no new exposure.

### Findings

- **Should-fix (1):** the churn cadence compares an all-phase pod count against the Deployment replica sum, so a lingering terminal-phase pod (e.g. an Evicted vLLM pod) pins the 5s cadence on permanently. Inline comment with the concrete scenario.

### Note, not a finding

- **Non-Deployment engines + scaled-to-zero → false drift signal.** The "intentionally scaled to zero → suppress message" logic keys off matching *Deployments* only (`desiredEngineReplicas`). Engines run as a StatefulSet (or bare pods) that are scaled to zero have no Deployment desired-replica source → `desiredKnown=false` → the controller emits `EngineSelectorUnmatched` + the drift message, indistinguishable from real drift. The message text ("no Pods in namespace match") is literally true, and Deployment-managed engines are the documented assumption, so this is a scope boundary rather than a bug — worth a one-line caveat in `cachebackend-api.md` if StatefulSet engines are in scope.

**Verdict:** approve-with-nits. The Should-fix is an efficiency/design-fidelity issue (functionally correct, just over-reconciles), not a correctness blocker.

internal/controller/cachebackend_controller.go
heymrbox: Addressed in 4cfa6f8.

I kept `status.matchedEnginePods` as the all-phase observed snapshot, but changed the churn cadence comparison to use only non-terminal matching pods (`Succeeded`/`Failed` are ignored for desired-vs-observed cadence). That keeps lingering Evicted/Failed pods visible to operators without pinning the reconciler to the 5s churn interval forever.

I added a regression test with two active matching pods plus one failed matching pod: the status count stays 3, while the requeue cadence stays at the steady 30s because the Deployment wants 2 live pods.

Verification:
- red/green focused regression test
- `go test ./internal/controller -run TestReconcileMatchedEnginePodsUsesSteadyCadenceWithTerminalPods`
- `go test ./internal/controller -run TestReconcileMatchedEnginePodsUses(ChurnCadenceWhenDeploymentDesiredDisagrees|SteadyCadenceWithTerminalPods|SteadyCadenceWithoutMatchingDeployment)`
- `go test ./internal/controller`
- `make pre-pr`

EdHasNoLife (approved):

PR #124, opened 2026-06-24
Add CachePolicy lookup strategy gates

heymrbox:
## Summary

Adds `CachePolicy.spec.strategy` to gate LookupRoute strategy behavior per namespace:

- `enableChainMatching` controls block-hash chain matching.
- `requireChain` rejects non-chain callers with `POLICY_REQUIRES_CHAIN`.
- `enableTenantHot` suppresses TENANT_HOT fallback when disabled.

Also wires the strategy through controller → server policy propagation, bumps `/policy` to v6 with backward-compatible defaults for older snapshots, updates CRD/generated artifacts, docs, samples, and unit/integration coverage.
## Linked issues

CAC-86

### Quality
- [x] CI is green.


heymrbox:
Addressed in 30f53fa.

What changed:
- Added `POLICY_REQUIRES_CHAIN` to the proto `LookupRouteResponse.reason_code` comment and regenerated the Go proto output.
- Moved `requireChain` enforcement to the wire request shape: the request must carry a valid `block_hashes` + `block_token_counts` chain. Legacy `prefix_hash`, malformed chains, `token_ids`, and `prompt_text` now return `POLICY_REQUIRES_CHAIN` before index lookup when the policy requires a chain.
- Restored the v5 rollout test to use a literal `version: 5` body, so v5 routing-floor opt-out plus v6 strategy defaulting stays pinned.
- Cleaned up the version-history wording and stale v5/v6 comments in the propagation docs/tests.

Verification:
- `make proto-lint`
- `go test ./pkg/server ./pkg/server/proto/inferencecache/v1alpha1`
- pre-push `make ci` gate passed before push

heymrbox:
Addressed in 773748b.

What changed:
- Added `POLICY_REQUIRES_CHAIN` to the top-level proto fail-open guarantee and regenerated the gRPC comments.
- Updated the server overview comment to list the policy-gated empty-result path.
- Tightened the gRPC contract doc so `requireChain` clearly requires a wire `block_hashes` + `block_token_counts` chain before dual-input fallback resolution.

Verification:
- `make proto-lint`
- `go test ./pkg/server ./pkg/server/proto/inferencecache/v1alpha1`
- pre-push `make ci` gate passed before push

heymrbox:
Addressed in 1b08e36.

What changed:
- Made `enableChainMatching=false` ignore malformed chain fields before chain validation, so a request with `prefix_hash` still uses the legacy exact path.
- When chain matching is disabled, well-formed explicit chains plus `token_ids` / `prompt_text` inputs now use the final block hash as the exact fallback key instead of sending an empty exact key to the index.
- Moved the `minimumPrefixTokens` check onto the effective lookup input, so ignored chain counts no longer satisfy the gate for an exact-prefix request.
- Added regression coverage for malformed-chain-plus-prefix, chain-only, malformed-chain-plus-token IDs, token IDs, prompt text, and the min-prefix ignored-count case.

Verification:
- `go test ./pkg/server`
- pre-push `make ci` gate passed before push

heymrbox:
Addressed in dc1e72e.

What changed:
- `enableChainMatching=false` now truly ignores chain-only requests instead of converting the final block hash into an exact lookup key.
- The disabled-chain regression test now asserts chain-only requests fail open with `NO_HINT` without touching the index.
- Mixed malformed-chain plus `token_ids` still uses the token-derived exact fallback, so lower-precedence non-chain inputs continue to work when chain fields are ignored.

Verification:
- `go test ./pkg/server`
- pre-push `make ci` gate passed before push

heymrbox:
Addressed in b3bc8d6.

What changed:
- Updated `docs/reference/reason-codes.md` so `TENANT_HOT` documents the `enableTenantHot` policy gate and `NO_HINT` documents the downgrade when tenant-hot is disabled.
- Refreshed the CachePolicy defaulter comment so it no longer claims eviction is the only marker-applied default.

Verification:
- `go test ./internal/webhook/v1alpha1`
- pre-push `make ci` gate passed before push

EdHasNoLife (approved):

PR #130, opened 2026-06-30
feat: events-only (tier-1 routing) integration mode for hybrid-attention models

EdHasNoLife:
Adds `spec.integration.mode: EventsOnly` (default `Offload`). Events-only wires an engine for cache-aware **routing** with no KV connector and no provisioned backend server — the supported integration for hybrid-attention models (Qwen3.6/Next gated-DeltaNet, Mamba/Jamba, …) that can't load a vLLM KV connector alongside their hybrid KV-cache manager.

- **Adapter:** events-only injects no connector; the subscriber forwards evictions (omits `--ignore-block-removed`, since no L2 tier retains blocks).
- **Pod webhook:** bypasses the `status.endpoint` injection gate so the subscriber is injected with no provisioned endpoint.
- **Controller:** `reconcileEventsOnly` — no server/Service, runs the KV-event readiness gate (Ready on first event) + matchedEnginePods.
- **Admission:** rejects events-only with `External` (contradiction) or `autoscaling` (no workload).
- Tests (adapter / webhook / validator / controller envtest), docs, sample.


EdHasNoLife:
Reviewer note — two advisory (Codex) items from the recent review were intentionally **not** actioned; flagging so they are not read as oversights:

- **Pod webhook — events-only "re-admission not idempotent" (subscriber strip).** The mutating pod webhook is `verbs=create` only, so the "first pass appends, second pass strips" sequence is not reachable for a single pod object. Leaving a pre-existing (operator-authored, therefore unverified) subscriber *unclaimed* on events-only is the documented, deliberate behavior — reversing it would reintroduce the forged `injected-by` stamp that an earlier commit here closed.

- **Controller — `External` + `EventsOnly` selectability predicate.** Admission already rejects the `type=External` + `mode=EventsOnly` pair, so an "active routing backend with no sidecar" is only reachable for an admission-bypassed / pre-existing stored object. Left as-is rather than adding a controller-side guard for a combination the validating webhook forbids.

Everything else from the review is addressed (mode-aware eviction docs sync, events-only stripping a stale kernel-check init container, and the diagnostic-message fix).

heymrbox (approved):

PR #131, opened 2026-06-30
feat: vLLM+Mooncake runtime adapter (managed Mooncake backend)

EdHasNoLife:
Adds a second `KVCacheRuntimeAdapter` for `(vLLM, Mooncake)` so an operator can apply a `CacheBackend{type: Mooncake}` and have engine pods auto-wired to a managed Mooncake store.

**Adapter** (`pkg/adapters/runtime/vllm_mooncake.go`)
- `ResolveCacheServer` renders a standalone `mooncake_master` Deployment + Service (RPC `:50051` first → `status.endpoint`; embedded HTTP metadata `:8080`).
- `InjectEngineConfig` merges (never clobbers) the engine wire onto the vLLM container.
- `ObservationSidecar` reuses the shared kvevent-subscriber (KV events are engine-side, `hash-scheme=vllm`).
- `Supports`/`SupportedPairs`/`ReservedArgs`/`ReservedEnv`/`EngineContainerName` implemented; registered in `DefaultRegistry` → admission admits the pair, C2 reconciles it managed, the pod webhook injects it.

**Design: Mooncake as an LMCache remote backend.** The engine runs the same `LMCacheConnectorV1`, pointed at `LMCACHE_REMOTE_URL=mooncakestore://<master>` (the `lm://` analog). vLLM's native `MooncakeStoreConnector` is configured only via a `MOONCAKE_CONFIG_PATH` file, which the env-only pod webhook can't write — routing the controller-resolved master address through the remote-URL scheme is the only path that reaches the engine via injection, and it matches the locked design that Mooncake "fits the `lm://`-style RemoteBackend wire". So the injected wire is byte-identical to LMCache except the URL scheme; the LMCache-connector core + subscriber builder are now shared.

**Real values pinned:** image `kvcacheai/mooncake:0.3.11.post1` (only published tag, version-aligned with `mooncake-transfer-engine` on PyPI); master cmd, RPC/metadata ports from upstream Mooncake docs.

**Scoped follow-ups (not invented):** transfer-engine `extra_config` (metadata/protocol/segment sizes) is operator-supplied via a config file, not env-injectable; a kind reference stack to validate the end-to-end manifest shape + digest-pin the image is tracked (TODO in code + design doc).

**Tests/docs:** unit tests for all four interface methods (merge-not-clobber, idempotency, sidecar, reserved sets) + direct `enginewire` tests; fake-client + envtest controller reconcile; admission-admits test; `config/samples/cachebackend-mooncake.yaml` (passes `make verify-samples`); `docs/design/cachebackend-api.md` + `kvevent-subscriber-wiring.md` updated.


EdHasNoLife:
Re: the recurring `vllm_mooncake.go` default-image Should-fix (Codex rounds 1–3) — disposition + rationale, since it is the one item kept open deliberately:

**Verification done:** the image `kvcacheai/mooncake:0.3.11.post1` and its linux/amd64 + linux/arm64 digests are confirmed published on Docker Hub and version-aligned with the `mooncake-transfer-engine` 0.3.11.post1 PyPI release (queried via the registry v2 API). I could **not** layer-inspect the config blob (entrypoint/PATH/ports) from this environment: the blob CDN sits behind a TLS-intercepting proxy here, and the downloaded config blob’s SHA-256 does not match the digest from the (trusted, non-intercepted) manifest chain — i.e. the content over that path is not authentic, so reading an entrypoint from it would be meaningless.

**Why deferred rather than blocking:**
1. **Fails safe + loud, not silent** — a wrong image/entrypoint/port means the `mooncake-master` pod CrashLoops/ImagePullBackOffs, the Deployment never reports Available, and the RPC-port readiness probe holds the CacheBackend at `Ready=False` (RolloutInProgress/ReplicasUnavailable). The operator sees it in `kubectl get cachebackend`, not a green-but-dead backend.
2. **Fail-open** — the cache is an optimization, so a broken master is never a serving outage; engines fall back to local prefill.
3. **Repointable** — `backendConfig.serverImage` overrides the default with no code change.
4. **Flagged everywhere** — godoc on `defaultMooncakeMasterImage`, the sample, and the API doc all mark it pinned-but-not-cluster-validated.

The proper close is the **A2-equivalent Mooncake kind reference stack** (the analog of the LMCache reference stack) — it would digest-pin the image and validate the master command/port surface end-to-end. That is scoped as the tracked follow-up; happy to gate a production default-flip behind it.

EdHasNoLife:
Status / disposition of the latest Codex review (verdict: changes-requested, but **0 Blocking, 0 Nit** — every category Clean except one Should-fix).

**All other findings across the review rounds are addressed and pushed** — the reserved-override admission test for the (vLLM, Mooncake) pair, the pod-webhook injection test, the reserved-declarations + resources + subscriber-wiring doc sync, the AIBrix migration of the "unsupported type" examples, the install-smoke Mooncake assertion, and the tracked design doc for the LMCache-vs-Mooncake persistence decision.

**The single remaining item is the default-image verification** (`vllm_mooncake.go`): the code honestly marks the `kvcacheai/mooncake:0.3.11.post1` entrypoint/ports as not independently validated, and the install-smoke uses a busybox stand-in rather than pulling the real image. I re-attempted to verify it via the registry config blob and the blob CDN is unreachable/untrusted from this build environment (TLS-intercepting proxy — the downloaded content’s SHA-256 does not match the trusted manifest digest), so the real entrypoint cannot be inspected here. The image + multi-arch digests are confirmed published and version-aligned with the `mooncake-transfer-engine` PyPI release.

This stays a **non-Blocking, deferred** Should-fix because the default fails **safe and loud**, not silently:
- a wrong image ⇒ the `mooncake_master` pod CrashLoops ⇒ the Deployment never reports Available ⇒ the RPC-port readiness probe holds the CacheBackend at `Ready=False` (visible in `kubectl get cachebackend`), not a green-but-dead backend;
- the cache is **fail-open**, so a broken master is never a serving outage;
- it is repointable via `backendConfig.serverImage` with no code change.

The real end-to-end validation (hand-built vLLM + Mooncake on kind, image digest-pin, and a real-engine `mooncakestore://` cache-reuse canary) is scoped as a tracked reference-stack follow-up — the analog of the LMCache reference stack. CI is green (incl. the install-smoke Mooncake assertion) and the PR is mergeable.

EdHasNoLife:
Update after syncing with `main` (the events-only routing mode #130 and EngineCompatibility #128 merged) — responding to the latest Codex review:

- **Merge conflicts resolved.** The one semantic conflict: #130 made the kvevent-subscriber `--ignore-block-removed` flag *integration-mode-aware*. That is now integrated into the shared `buildKVEventSubscriber`, so both the LMCache and Mooncake adapters honor it — `Offload` (incl. Mooncake, always an L2 store) sets the flag; `EventsOnly` omits it (no L2 tier, so `BlockRemoved` is a real eviction).
- **Closed a gap the merge exposed.** With the `(vLLM, Mooncake)` adapter now registered, `spec.type: Mooncake` + `mode: EventsOnly` became admittable but is contradictory (events-only wires no KV connector, so a `mooncake_master` store would be provisioned with nothing to use it). Admission now enforces **EventsOnly requires `spec.type: LMCache`** (`rejectEventsOnlyMisconfiguration`, with a test); the design docs state the same. Before the Mooncake adapter shipped this was implicitly caught by the runtime-adapter check.
- **All other findings across the review rounds are addressed** (doc sync, the `--ignore-block-removed` per-mode contract, the podinjector managed-types comment, the smoke tunables).

The **only** remaining Should-fix is the default-image entrypoint/port verification (0 Blocking, 0 Nit) — unchanged rationale: the image + digests are confirmed published, the value is not layer-inspectable in this build environment, and it fails **safe + loud** (`Ready=False`) + fail-open + is repointable via `backendConfig.serverImage`. Real-image validation + digest-pin is scoped to the reference-stack follow-up. CI is green (13/13, incl. the install-smoke Mooncake assertion) and the PR is up to date with `main` and mergeable.

heymrbox (approved):

PR #141, opened 2026-07-10
fix(mooncake): provision the master on hostNetwork behind a headless Service

EdHasNoLife:
A managed Mooncake backend reconciles `Ready` and transfers **zero KV**, silently. Validated on a real cluster — the master's key count never left 0.

**Why.** Mooncake is a peer-to-peer transfer-engine **mesh**, not a single-endpoint server like LMCache's `lm://`. The master on `:50051` returns only a directory pointer ("this block lives on node B"); the engine then dials that node's real IP on a **dynamically negotiated port**. A ClusterIP Service forwards only the ports declared on it, and CNI overlay pod IPs are unreachable for the mesh.

**Adapter**
- master pod: `hostNetwork` + `dnsPolicy=ClusterFirstWithHostNet`
- Service: **headless** (`clusterIP: None`) — its DNS name is already what `serviceEndpoint` publishes into `status.endpoint`, so it now resolves straight to the master's node IP with every port reachable. **No endpoint-resolution change needed**, and the endpoint survives a reschedule.

**Reconciler** (each of these was a real, separately-caught bug)
- `Recreate` strategy for hostNetwork pods — a rolling surge collides on the node's ports (the apiserver defaults `hostPort=containerPort` there).
- Propagate `spec.clusterIP` on Service **create** — the mutate func never set it, so a headless render silently got an allocated virtual IP.
- **Recreate the Service** when headless-ness diverges — `clusterIP` is immutable, so an in-place migration fails forever and leaves the backend silently broken.
- Reconcile `hostNetwork`/`dnsPolicy` **and** the strategy on **update** — `applyDeployment` only overwrites Spec on create, so an already-provisioned backend would never migrate and the fix would be a no-op on upgrade.

**Blast radius.** `type: LMCache` is unaffected (pod network + virtual ClusterIP), guarded by tests. Engine-side hostNetwork injection is a separate change.

**Tests.** Adapter unit; `buildDeployment` strategy + `headlessnessDiverges` unit; **envtest** against a real apiserver covering clusterIP allocation, immutability, and both upgrade migrations — each assertion **mutation-verified** to fail without its fix; plus a default-install **smoke** assertion (hostNetwork + Recreate + headless). Docs: operator-facing hostNetwork requirement + trade-offs.

No proto/CRD schema change.


EdHasNoLife:
Thanks — all three actionable findings addressed in a27b937. One pushback, with evidence.

**Blocking (`vllm_mooncake.go:274`) — fixed, and you were right.** `Recreate` only serializes rollouts; it does nothing about simultaneous replicas. The deeper problem is that the Mooncake master is a **singleton**: a second replica either fails to schedule or comes up as an *independent* master and silently splits the store. Rather than paper over it with `hostPort` + scheduling constraints (which would only turn a split store into an unschedulable pod), admission now **hard-rejects** `spec.replicas > 1` and `spec.autoscaling` for `type: Mooncake` (`rejectMooncakeMasterScaleOut`). `replicas: 0` (disabled) and `1` stay valid; `type: LMCache` scales exactly as before (guarded by a test).

**Should-fix ×2 (`cachebackend_controller.go:1119`, `:1279`) — fixed.** Both now compare against the **effective** desired value (an unset adapter value means "the API-server default"), so the migration is symmetric: switching away from Mooncake restores `RollingUpdate` and `ClusterFirst` instead of stranding the Deployment. Re-writing an already-correct value is skipped, so a pod-network backend still never churns. Covered by a new envtest (`RevertsToPodNetworkAndRollingUpdateWhenTypeChangesAwayFromMooncake`).

**Nit (`:986`) — respectfully disagree; the comment is accurate.** Kubernetes *does* default `hostPort = containerPort` for host-network pods — `defaultHostNetworkPorts` in the core/v1 defaulting (`SetDefaults_Pod`), applied when `spec.hostNetwork` is true. Verified against a real apiserver (envtest), creating pods that declare **only** `containerPort`:

```
HOSTNETWORK pod: containerPort=50051 hostPort=50051   <- defaulted
POD-NETWORK pod: containerPort=50051 hostPort=0
```

That defaulting is exactly why the scheduler's `NodePorts` predicate rejects a second same-node master (observed on a live cluster: `0/25 nodes are available: 1 Insufficient ... Predicate NodePorts failed`), which is why `Recreate` is required. Leaving the comments as-is, since the claim they make is the load-bearing reason for the strategy.

All prior assertions plus the new ones are mutation-verified — each fix was removed in turn and the test confirmed to fail with the intended message before restoring.

EdHasNoLife:
Blocking addressed in the latest push — thank you, this was the right call to hold the line on.

You are correct that the managed path is still incomplete: the master is now provisioned on the host network, but Mooncake's transfer engine is a peer-to-peer mesh and the **engine** pods must run with `hostNetwork` too, which the pod webhook does not inject yet. Engine-side injection is a deliberately separate change (it mutates a user-owned workload, so the inject-vs-opt-in policy needs its own review).

What I have **not** done is leave that gap silent. Admission now emits a **warning on every `type: Mooncake` create and update**:

> `spec.type=Mooncake: the master is provisioned on the host network, but engine pods must ALSO run with hostNetwork for Mooncake's peer-to-peer transfer engine to reach it — the pod webhook does not inject that yet. Until you set hostNetwork on your engine workload, this backend will report Ready while transferring zero KV. The namespace must also permit hostNetwork (Pod Security 'restricted' rejects it).`

A warning rather than a rejection, because the backend *is* admissible and the master half genuinely works — rejecting it would block the very operators who are about to set `hostNetwork` on their engine. The exact failure you named ("healthy-looking backend that transfers zero KV") is now stated at apply time, in the operator's terminal.

Guarded three ways so it cannot rot:
- unit tests: warning present on create **and** update; absent for `type: LMCache`
- **default-install smoke** asserts the real `kubectl apply` prints it
- both mutation-verified (removed the warning; confirmed the tests fail)

Also updated `cachebackend-api.md` to name this as the one gap the managed path does not close for you yet.

Re: the two Should-fix items and the Nit on `hostPort` defaulting — I have to respectfully hold. Kubernetes *does* populate `hostPort` from `containerPort` for host-network pods (`defaultHostNetworkPorts`, invoked from `SetDefaults_Pod`). Verified against a real apiserver:

```
HOSTNETWORK pod: containerPort=50051 hostPort=50051   <- defaulted
POD-NETWORK pod: containerPort=50051 hostPort=0
```

and observed on a live cluster as `Predicate NodePorts failed` when a second master was scheduled. That said, I took the substantive point: the messages and comments now **lead with the failure an operator actually sees** — a second master cannot bind ports the first holds, and on another node it becomes an independent master that splits the store — with the scheduler predicate described as the earlier, secondary rejection. So the wording is correct regardless of how a given cluster resolves it.

EdHasNoLife:
Both addressed in the latest push — good catches, and the first one was a real hole in my own test.

**Should-fix (reverse Service migration) — fixed.** You are right: `RevertsToPodNetwork…` asserted only the Deployment, so the headless → virtual-ClusterIP direction was never exercised end-to-end. That is the same immutable-field problem in reverse, and it is exactly the case a fake client cannot cover (it neither rejects the in-place update nor allocates the replacement VIP). The test now asserts the divergent Service is deleted on the switching reconcile and that the next one recreates it with an apiserver-allocated virtual IP. I verified the assertion bites: making `headlessnessDiverges` one-way (`live != None && desired == None`) fails it.

**Nit (warning length) — fixed.** Trimmed 450 → **116 chars**:

> `Mooncake: engine pods must also set hostNetwork (not auto-injected); without it the backend is Ready but moves no KV`

The full rationale (the mesh, the node-port reservation, Pod Security) already lives in `cachebackend-api.md`, which is the right home for it. Added `maxWarningLen = 120` plus a test asserting the constant stays inside it — a truncated warning is precisely the silent failure the warning exists to prevent, so the budget is now guarded rather than left to review. The install-smoke assertion was updated to the new text, so it still proves the warning reaches a real `kubectl apply`.

EdHasNoLife:
All three addressed in the latest push. The Blocking was a real hole, and it was mine — thank you for finding it.

**Blocking (grandfathered scale-out) — fixed in the reconciler, which is the only place that can fix it.** You are exactly right: `ValidateUpdate` filters to violations an edit *introduces*, so a CacheBackend written before the rule (or before its backend moved onto host networking) keeps `replicas: 3` + an HPA and is never re-validated. Worse, **my change is what makes that state dangerous** — those replicas used to be merely useless behind a ClusterIP; on host networking they contend for node ports or become independent masters that split the store. Admission cannot retroactively remediate an object already in etcd, so:

- `clampHostNetworkToSingleton` caps a host-network Deployment at one replica and runs **last**, after both the desired spec and the HPA-preserved live value. `0` is preserved (that is "disabled", not "scaled out").
- `reconcileHPA` now treats a host-network pod as autoscaling-disabled and tears down any HPA it owns — decided from the **observed pod shape**, not the spec, so a grandfathered HPA cannot undo the clamp on its next scaling decision.

Covered by an envtest that simulates precisely the grandfathered object (`replicas=3` + `spec.autoscaling` + an HPA-scaled live Deployment; the apiserver would now refuse to persist one, so it is constructed in memory). Both the clamp and the HPA teardown are mutation-verified — disabling either fails the test.

**Should-fix (network exposure) — documented.** Added an explicit caveat: host networking publishes the master's RPC/metadata/metrics ports **and the transfer engine's dynamically negotiated data ports** on the node's interfaces, and `NetworkPolicy` selects pods by pod IP so it **does not constrain a hostNetwork pod's listeners** — the pod-network isolation operators expect is simply absent. Isolation must come from node-level controls (security groups / firewall rules, and constraining which nodes may host the master and its engines). None of those ports authenticate callers.

**Nits (stale comments) — fixed.** `applyDeployment`'s ownership description now names strategy, host networking, DNS policy, volumes, and pod overrides (and the singleton clamp's precedence over the HPA), and `reconcileManagedPodSpec` no longer lists `DNSPolicy` among the fields it leaves untouched.

heymrbox (approved):

PR #142, opened 2026-07-10
feat(mooncake): opt engine pods into host networking via spec.integration.engineHostNetwork

EdHasNoLife:
Stacked on #141 (`mooncake-master-hostnetwork`) — review that first; this diff is only the engine side.

**Problem.** Mooncake is a peer-to-peer transfer-engine mesh, not an `lm://`-style single-endpoint server. The master returns a directory pointer; the engine then dials a real node IP on a dynamically negotiated port. A CNI overlay pod IP cannot. #141 fixed the master; an overlay *engine* pod still leaves the backend `Ready` with **zero KV transferred**.

**Change.**
- New optional `spec.integration.engineHostNetwork`. When set, the Pod webhook moves matched engine pods onto the host network (`hostNetwork` + `dnsPolicy: ClusterFirstWithHostNet` — `status.endpoint` is a Service DNS name, so cluster DNS must survive).
- **Opt-in, not default.** It rewrites the networking of a pod the *operator* owns. `hostNetwork` is a privilege, and mutating webhooks run **before** Pod Security validation — silently injecting it would turn a working engine pod into one a `restricted` namespace rejects, with an error naming Pod Security rather than this controller.
- Admission **rejects** the flag on backend types that don't need it, so it can't sit inert and read like a granted privilege. Covers the update path too (retyping away from Mooncake would otherwise carry it along as a no-op).
- The apply-time warning now names the exact field, and goes quiet on opt-in.

**Verification.** Every new assertion mutation-verified.
- Pod webhook, full admission path (not just the adapter): Mooncake injects only on opt-in; LMCache never injects.
- Validator: reject on non-Mooncake at create *and* on the type flip at update; dropping the flag with the flip is accepted (the rule must not wedge the object).
- Install smoke asserts both directions — warning present without the opt-in, absent with it.
- `make ci` green; `make verify-naming` clean.

Field-level rationale lives in the CRD godoc and `docs/design/cachebackend-api.md`; the sample is updated.


EdHasNoLife:
Thanks — pushed `ff17aba`. Adopted both Should-fixes and the docs point from the earlier pass. On the Blocking item I'm **declining the proposed remedy**, with evidence, but I've pinned the concern as a regression test.

### Blocking: "hoist the hostNetwork mutation above the endpoint gate"

Both halves of the premise turn out not to hold. I probed each against the real code rather than reasoning from the diff:

**1. `EventsOnly` is unreachable here.** `rejectEventsOnlyMisconfiguration` (`cachebackend_webhook.go:978-988`) forbids `EventsOnly` on every managed type except LMCache, and `engineHostNetwork` is rejected on every type except Mooncake. The two rules cross, so `Mooncake + EventsOnly + engineHostNetwork` cannot be created:

```
Mooncake + EventsOnly + engineHostNetwork => err=spec.integration.mode: Forbidden:
  mode "EventsOnly" is only supported with spec.type "LMCache"; got spec.type "Mooncake"
```

**2. The endpoint-not-ready pod is not "wired but unprivileged" — it is entirely un-wired.** `Handle` returns `failOpen` before any injection, and `failOpen` returns `admission.Allowed(reason)` with **no patches**:

```
Allowed=true patches=0 msg="CacheBackend status.endpoint not usable (fail-open)"
RESULT: hostNetwork=false  hasRemoteURL=false  dnsPolicy=""
```

So there is no state in which a pod dials the Mooncake mesh but lacks host networking. `hostNetwork` and `LMCACHE_REMOTE_URL` are injected together, behind the same gate — and `InjectEngineConfig` returns on the wire-injection error *before* touching `hostNetwork`, so even bypassing the gate fails safe (`inject engine config: endpoint is empty`, patches=0).

Hoisting the mutation above the gate would make it *worse*: it grants `hostNetwork` — a privilege Pod Security gates — to a pod that receives no connector and therefore has nothing to use it for. And it doesn't buy anything, because a Pod's `hostNetwork` is immutable: that pod must be recreated once the backend reports `Ready` in any case, since it is missing the connector env too. On recreation it gets both.

The coupling is deliberate, so it is now **pinned** rather than merely argued — `TestHandle_MooncakeBackend_HostNetworkNeverGrantedToAnUnwiredPod`. I verified it fails against a faithful implementation of the suggested change (one that patches `hostNetwork` onto the un-wired pod):

```
--- FAIL: TestHandle_MooncakeBackend_HostNetworkNeverGrantedToAnUnwiredPod
    pod was granted hostNetwork while the backend had no endpoint — a
    Pod-Security-gated privilege handed to a pod with no KV connector to use it
```

Also added `TestValidator_EngineHostNetworkCannotGoInertViaEventsOnly`, since that safety is emergent from two independent rules meeting and would vanish silently if either were loosened.

### Should-fix: docs must say existing pods aren't moved — **adopted**
`cachebackend-api.md` now states that injection happens at admission, that `hostNetwork` is immutable on a Pod, and that the engine workload must be rolled after enabling the flag.

### Should-fix: smoke doesn't exercise the pod mutation — **adopted**
The install smoke now admits a matching engine pod through the *installed* webhook and asserts `hostNetwork: true`, `dnsPolicy: ClusterFirstWithHostNet`, and `LMCACHE_REMOTE_URL=mooncakestore://<endpoint>` on the persisted object. It also asserts a **genuinely injected** non-Mooncake engine pod stays on the pod network — placed where that backend is still live, so an empty `hostNetwork` there means "the webhook left it alone", not "the check silently skipped".

### Should-fix (earlier pass): engine scheduling/rollout constraints — **adopted**
Documented that the API server defaults `hostPort=containerPort` for hostNetwork pods, so there is at most one engine replica per node per serving port, and that a `RollingUpdate` engine Deployment deadlocks (the surge pod cannot bind a port the outgoing pod still holds) — use `Recreate` / `maxSurge: 0`. Called out explicitly that this lands on the operator's Deployment, which the controller does not own and cannot clamp the way it clamps the master.

Note CI: `ci.yml` is gated on `branches: [main]`, so it does not run on this stacked PR. I ran `make ci`, `make verify-naming` and `govulncheck` locally (all green), and dispatched `default-install-smoke.yml` against this branch's SHA.

heymrbox (approved):

PR #146, opened 2026-07-14
docs: LMCache MP mode — converged worker model (SGLang phase 1)

EdHasNoLife:
Stacked on #145 (rebase onto main once #145 merges). Phase 1 of the SGLang+LMCache adapter fix — design + a stale-comment resolution, **no behavior change**.

- New `docs/design/sglang-lmcache-mp-mode.md`: the GPU-validated MP-mode design that replaces the shipped "mirror the vLLM `lm://` adapter" model.
  - SGLang drives LMCache in **MP mode** (node-local worker + `--lmcache-config-file`, `LMCACHE_*` env ignored); cross-node sharing is a **networked L2** (`resp`/Redis, proven) behind the worker, not `lm://`.
  - Fits the existing `KVCacheRuntimeAdapter` with **no new methods**: `ResolveCacheServer` → Redis L2; `InjectEngineConfig` → config-file init container + MP-worker sidecar + engine wire. **No engine `hostNetwork`** (worker is a same-pod sidecar) — the key divergence from Mooncake.
  - Includes the live node-local + cross-node proofs (`0 L1, 14 L2`).
- `enginewire.go`: resolve the stale `TODO(wire-test before production)` in `InjectSGLangLMCache` — validation answered it (MP mode; injected env inert). Comment-only.

Phase 2 = the working data plane (Redis L2 + worker sidecar + config-file wire; resolve the GPU-less-sidecar spike; flip the advisory warning off). Phase 3 = samples/reference-stack/smoke/docs.

Note: base is a non-`main` branch, so CI runs Codex only until retargeted to `main`.

PR #147, opened 2026-07-14
feat(sglang): render the shared Redis L2 store for MP mode (phase 2, increment 1)

EdHasNoLife:
Increment 1 of the SGLang LMCache MP-mode data plane (design: #146). Independent of #146 — touches no shared code. **Merge after #146** (a godoc reference to the design doc resolves once #146 lands).

- Adds `ResolveRedisL2Server` (`pkg/adapters/runtime/redis_l2.go`): the shared **L2 store** the SGLang MP worker offloads to via `--l2-adapter resp`. `lm://` is not a valid MP `--l2-adapter` type, so the SGLang pair can't reuse `ResolveLMCacheServer`.
- Redis for the default L2: fits the **ClusterIP, engines-anywhere** model (no `hostNetwork`/mesh), one Deployment+Service, vendor-neutral, proven end-to-end. `s3`/`mooncake_store` are durability opt-ins for later (BYO via `backendConfig.l2Adapter`, wired in increment 2).
- Ephemeral (`--save ""`, `--appendonly no`); `--maxmemory` = 80% of `spec.resources` memory (`base - base/5`: overflow-safe, inherently positive so never `0`/unlimited) + `allkeys-lru`, so LRU — not the OOM killer — reclaims. Versioned default image (production **must** digest-pin via `backendConfig.redisImage`), TCP readiness, ClusterIP.
- Security posture: unauthenticated ClusterIP holding KV blocks — same as the `lm://` server it replaces; NetworkPolicy/AUTH/TLS is follow-up hardening.
- Unit tests: render shape (image/pull-policy/ports/probe/service), image override, maxmemory derivation + overflow/positivity boundaries + fits-within-limit, ClusterIP-only, nil.

**Not wired into the adapter yet** — exported + tested on its own so the store render is reviewed apart from the engine wire. Increment 2 (`ResolveCacheServer`→Redis + MP-worker sidecar + config-file wire + `l2Adapter` BYO knob) stacks on #146 and does the wiring; `backendConfig.redisImage` gets documented in `cachebackend-api.md` then (it has no effect until wired).


EdHasNoLife:
Pushed fixes for the Codex should-fixes. Two are intentional deferrals, not oversights:

- **design-doc reference** (`redis_l2.go` godoc → `sglang-lmcache-mp-mode.md`): that doc is added by #146. #147 should merge **after** #146 (rebasing onto main then resolves the reference). Both are in flight.
- **`backendConfig.redisImage` api-doc sync**: lands with the SGLang `cachebackend-api.md` rewrite in increment 3 — the key isn't consumed until increment 2 wires this render, so documenting it before it does anything would be premature.

The singleton-enforcement + security-posture concerns are now addressed in the godoc (enforcement stated as a wiring prerequisite; security posture documented = same as the lm:// server it replaces).

EdHasNoLife:
Round: 3 nits fixed in 8bbf9ee.

- **Inverted headroom claim** (`redis_l2.go`, test) — correct catch: rounding `base - base/5` UP yields a *larger* budget, i.e. marginally *less* cgroup headroom, not more. I had the direction backwards. Reworded; the direction is still safe because the result never reaches `base`.
- **"exactly 80%"** (`cachebackend-api.md`) — now "~80% (integer arithmetic, rounds up)" at both mentions, consistent with the code comment.

heymrbox (approved):

PR #149, opened 2026-07-16
feat(sglang): LMCache MP-mode engine wire — the working data plane (phase 2, increment 2)

EdHasNoLife:
Increment 2 of the SGLang LMCache MP-mode data plane — the wire that makes `(sglang, LMCache)` actually cache. **GPU-validated end-to-end** with the exact adapter-rendered shape (store→flush→retrieve reused 3072 tokens after a GPU-cache flush).

Stacked: #146 (design) → #147 (Redis render) → this. **Merge order matters** — merge #146, then #147; this then rebases onto `main` (the one `enginewire.go` godoc overlap resolves in favor of the rewrite).

- **`ResolveCacheServer` → Redis L2** (via #147's render), not the dead `lm://` server.
- **`InjectSGLangLMCache` rewritten** (`sglang_mp.go`): a node-local **MP-worker native sidecar** (`restartPolicy: Always` init container) that writes the config file then runs the LMCache MP server; `NVIDIA_VISIBLE_DEVICES=all` so the GPU-less sidecar CUDA-IPCs the engine's GPU with no device allocation; **exec** startup-probe on the loopback ZMQ port gates the engine start; shared config + `/dev/shm` volumes (an existing engine `/dev/shm` mount is **reused**, since a duplicate `mountPath` is an invalid Pod); engine gets `--enable-lmcache` + `--lmcache-config-file` + `LMCACHE_USE_EXPERIMENTAL`, and the inert `LMCACHE_REMOTE_URL`/serde/local-CPU env is **dropped**.
- **Sanitized tunables**: `chunkSize`/`mpPort`/`l1SizeGB` are substituted into the worker's `sh -c` command, so they're constrained to in-range positive integers (injection guard + a bounded `/dev/shm`).
- Reserve `--lmcache-config-file`; drop `LMCACHE_REMOTE_URL` from `ReservedEnv`.
- **Remove the obsolete "offload misconfigured" admission warning** (+ tests + install-smoke assertion, now flipped to require a clean admit); retire the KNOWN LIMITATION godocs; rewrite the `cachebackend-api.md` SGLang contract to the MP wire.
- Tests rewritten for the MP wire (sglang / podinjector / validator), incl. sanitization, bounds, and `/dev/shm`-reuse regression guards. Whole suite green.

**Deferred:** the BYO `l2Adapter` knob (operator-supplied external L2) is **not** in this PR — skipping the managed Redis clears `status.endpoint` and the webhook's empty-endpoint gate then skips injection entirely. Tracked as a follow-up; see the deferral commit + [the PR comment](https://github.com/cachebox-project/inference-cache/pull/149#issuecomment-4994981639) for the root cause and the chosen fix (an adapter-aware gate).

Follow-up increment: samples, reference-stack leg, `VERSIONS.md` pin, api-doc polish.


EdHasNoLife:
**BYO `l2Adapter` is deferred out of this PR** (see the `defer the BYO l2Adapter knob` commit). Tracked as a follow-up; this PR ships exactly the GPU-validated managed-Redis path.

**Why it was pulled:** skipping the managed Redis makes `ResolveCacheServer` return no server → the reconciler reconciles the backend unmanaged and **clears `status.endpoint`** → the pod webhook's **empty-endpoint gate** then skips engine injection entirely, so a BYO backend would cache **nothing** — worse than not having the knob.

**Root cause:** that gate (`internal/webhook/pod/podinjector.go`) exists *only* to protect vLLM's `lm://` dial target — its own comment says so: *"the connector requires a dial target (an empty/malformed LMCACHE_REMOTE_URL crashes the engine at startup)."* SGLang MP has no dial target: the engine dials its **node-local worker** at `127.0.0.1`, never `status.endpoint`, and the worker **fail-opens on an absent/unreachable L2** (GPU-validated — it comes up L1-only and retries). So the gate skips SGLang injection to prevent a failure mode SGLang doesn't have. It already bypasses for events-only for the same class of reason.

**Chosen fix (follow-up):** make the gate **adapter-aware** instead of special-casing SGLang in the webhook — add an `EngineWireNeedsEndpoint() bool` capability to `KVCacheRuntimeAdapter` (vLLM/Mooncake `true`, SGLang `false`) and have the webhook consult it, so the gate stops encoding a vLLM assumption. Then BYO can skip provisioning **and** still inject. Bonus: it also fixes a latent managed-path foot-gun — an SGLang engine pod created *before* the managed Redis is `Ready` currently boots un-wired.

EdHasNoLife:
Round 7 — all three code/doc findings fixed in 0480408; the fourth is a stacked-PR artifact.

- **resp port not validated** (`sglang_mp.go`) — real, fixed. The port is emitted unquoted into JSON, so a non-numeric/out-of-range one rendered invalid config: the worker never parses `--l2-adapter`, never binds ZMQ, and the engine sits behind the startup probe forever. Now rejected at admission (→ fail-open admit, pod starts un-wired), with coverage for non-numeric / >65535 / zero.
- **`l1SizeGB` "ignored"** — real, fixed, and my wording was the bug. Only the tmpfs sizing drops out when the engine brings its own `/dev/shm`; `--l1-size-gb` and the worker memory request+limit still follow the key. Table now says so explicitly.
- **`upsertContainerByName` / `upsertVolumeByName`** — real, fixed. Renamed to `adoptContainer` / `adoptVolume` in the round-6 fix; the doc row lagged.
- **`sglang-lmcache-mp-mode.md` missing** — expected, not fixable here. This PR is stacked on #147, and the design doc lands in #146 (still open). Merge order is #146 → #147 → #149; the link resolves when this rebases onto main, which is also when CI first runs (the workflow only triggers on PRs targeting `main`, so the gate on this stack is the local `make ci`, green on both round-6 and round-7 pushes).

EdHasNoLife:
Round 8 — both Blockings were real and are fixed in 38a38e9, with one caveat I want to flag rather than bury.

**Blocking: IPC_LOCK vs Pod Security** — real, and worse than stated: added capabilities beyond a small allow-list are disallowed under **baseline as well as restricted**, and IPC_LOCK is on neither list. Since this mutation lands before Pod Security admission, any enforcing namespace would reject the whole engine pod — the cache plane breaking the engine, which inverts the fail-open contract. Removed rather than made opt-in: it was carried over from the RDMA-oriented Mooncake reference manifests, the MP wire moves KV over CUDA-IPC + `/dev/shm` (not RDMA), and the *engine* side — which this adapter never grants capabilities to — would have needed it too if the data path did. A worker-only spike ran the MP server GPU-less with no securityContext at all.

⚠️ **Caveat: the removal is not yet re-validated on GPU.** The end-to-end run (`cached: 3072`) carried IPC_LOCK on both containers, so the reasoning above is argued, not measured. The cluster token has expired; I will re-run the adapter-shaped validation without the capability before this merges. Flagging explicitly so the GPU-validated claim in the design doc is not read as covering this change.

**Blocking: /dev/shm subPath** — real, fixed. Same volume is not the same directory: an engine mounting subPath `shm` against a worker mounting the volume root admits cleanly and then transfers zero KV. The worker now mirrors the engine mount's subPath. `subPathExpr` is rejected instead of mirrored — it expands `$(VAR)` from the mounting container's env, and the worker's env is not the engine's, so the same expression can resolve elsewhere or not at all.

**Should-fix: source-level readOnly** — fixed (`persistentVolumeClaim.readOnly`, `csi.readOnly` alongside the projection sources).

**Should-fix: forgeable ownership marker** — the value is now checked, not just the name. The deeper point stands and is documented as an accepted boundary: nothing in a PodSpec proves provenance, and `InjectEngineConfig` receives no object metadata to carry an authenticated marker. An operator who both names their container `lmcache-mp-worker` and stamps our exact marker has declared it adapter-owned. What the marker buys is the case that actually happens — an accidental collision carries no marker, so it is rejected, not overwritten.

**Nits** — both fixed: the doc now separates the two override surfaces (`engineOverrides` = engine args/env; `backendConfig` = worker tunables; neither reaches the other), and the smoke matches the retired warning's exact string rather than the prose fragments `misconfigured` / `MP mode`.

**`sglang-lmcache-mp-mode.md` missing** — unchanged from round 7: stacked-PR artifact, resolves when #146 merges and this rebases onto main.

EdHasNoLife:
Round 9 — nit fixed in f93fea3; the Blocking is accepted and being settled empirically rather than by argument.

**Blocking: `NVIDIA_VISIBLE_DEVICES=all`** — accepted as a real isolation break: with the NVIDIA container runtime this does expose every GPU on the node to a sidecar that holds no device-plugin allocation, so on a shared GPU node our injected worker can reach another workload's device.

I am not going to remove it on reasoning alone, because the same reasoning is what produced the IPC_LOCK caveat in round 8, and both are claims about what the MP data path actually requires. LMCache MP has two data paths — CUDA-IPC (`gpu`) and POSIX `/dev/shm` (`non_gpu`, via `--shm-name`) — and this adapter already renders the `/dev/shm` L1 tier, so the worker may not need GPU visibility at all. That is testable, not arguable.

Re-running the GPU validation on H100 now to settle both in one pass:
1. worker with `NVIDIA_VISIBLE_DEVICES` dropped — does store→flush→retrieve still reuse KV?
2. worker without `IPC_LOCK` (removed in 38a38e9, currently unvalidated) — same question.

Whatever the run says goes into the render and the design doc as evidence. If the worker does need GPU visibility, the honest outcome is to keep it and document the isolation trade-off explicitly for shared-node operators — the same "explicit opt-in category" framing this project already applied to Mooncake's `hostNetwork` — rather than ship a wire that silently caches nothing.

**Should-fix: smoke only proves the warning is absent, not that the wire is injected** — agreed, and it is the right assertion. Folding it into the follow-up increment that adds the reference-stack leg, for two reasons: that increment already owns the samples + stack legs, and the assertions depend on the GPU-visibility outcome above (writing them now risks pinning a render that is about to change).

**`sglang-lmcache-mp-mode.md` missing** — unchanged: stacked-PR artifact, resolves when #146 merges and this rebases onto main.

EdHasNoLife:
Round 9 Blocking — settled on the GPU, and the measurement went against my hypothesis. 8647e4e.

**Setup**: A100 node, llama-3.1-8b, 7007-token prompt, `/flush_cache` between passes so the GPU radix tree is cleared and any hit can only come from LMCache. Pod rendered by calling the *actual adapter* (`InjectEngineConfig`) rather than hand-written YAML — the previous hand-written manifest had silently drifted from the shipped render. Distinct prompt per variant so the shared Redis L2 could not let one variant false-hit on another (confirmed: each cold pass shows `cached_tokens=1`).

| variant | worker | result |
|---|---|---|
| v2 | IPC_LOCK removed, `NVIDIA_VISIBLE_DEVICES=all` | **cached 6912/7007** ✅ |
| v3 | IPC_LOCK removed, our NVIDIA env **dropped** | **cached 6912/7007** ✅ |
| v4 | IPC_LOCK removed, `NVIDIA_VISIBLE_DEVICES=void` | worker crash, **engine never ready** ❌ |

**IPC_LOCK**: removal is now measured, not argued. Round 8's caveat is discharged.

**GPU visibility: the finding is real, but the requested fix is not achievable — and I was wrong about the mechanism.** I claimed the worker might not need the GPU (the `non_gpu` /dev/shm path). It does: the logs show it registering the engine's KV tensors and opening a CUDA stream on the engine's device. Revoking visibility is fatal, not degrading:

```
RuntimeError: Device UUID 876eebab-... not found in the discovered devices.
              Please make sure the process can see all the accelerator devices
  .../lmcache/v1/platform/cuda/ipc_wrapper.py: _get_device_index_from_uuid
```

The engine hands the worker a device **UUID**; LMCache resolves it to a local index, which requires the device to be visible. Narrowing it to the engine's device is **not available to a mutating webhook** — the device plugin assigns the UUID at kubelet time, *after* this mutation runs, so there is nothing to narrow to. A `nvidia.com/gpu` request for the worker is strictly worse: it burns a second GPU and the scheduler hands it a *different* device than the engine's.

**And dropping our env var is not the lever it looks like** (v3): sglang images ship `NVIDIA_VISIBLE_DEVICES=all` in their own `ENV`, and the device plugin overrides it **only** for containers that request a GPU. The engine gets a UUID; the request-less worker keeps the image default. v3 saw all 8 GPUs with our env removed and cached identically — so removing it buys zero isolation while making the wire depend on an image side effect (a slim `workerImage` without that ENV would break).

So I took the second half of your ask — "an explicitly documented safe isolation mechanism" — and documented the trade-off rather than implying it away: an operator-facing note in `cachebackend-api.md` stating the worker sees every GPU on the node, why it cannot be narrowed, that it holds no device-plugin allocation, and that tenants needing hard GPU isolation should not co-schedule on one node. Same framing this project already applied to Mooncake's `hostNetwork`: a real constraint, stated plainly, not hidden. A test pins the env so it is not later dropped as dead weight — the failure that causes is a wedged engine, not a cache miss.

**Should-fix (smoke proves absence, not injection)** — agreed and deferred to the increment that adds the reference-stack leg, per the reasoning in my previous comment.

EdHasNoLife:
Round 11 — all three addressed in 56d2bbc. Two were real gaps the merged design doc made visible; the Blocking is a genuine design↔impl mismatch, deferred deliberately with the owner’s call.

**Blocking — worker has no liveness/self-healing.** Real: the design’s containment list promises four measures “all in Phase 2”, and this increment ships two (the native sidecar’s `restartPolicy: Always` self-heal, and docs naming the worker a serving component). The engine liveness probe and the `CacheBackend` `Degraded` condition are **deferred to Phase 2’s increment 3**, and I have written that into the design doc rather than leaving the promise floating — see the new increment table under [Phased delivery] and the annotated containment list.

Why deferred rather than added here, in one line each:
- They are **different surfaces**. `Degraded` is controller/status work; an engine liveness probe mutates the **operator-owned engine container**, which may already carry its own probe. Neither is the engine wire this PR renders.
- Both are **gated on a Phase-2 finding this design explicitly does not claim** — the “Worker crash / restart” bullet: *whether the engine survives a mid-flight worker restart*. A restart-on-unhealthy loop built before that is measured could convert “caching silently stopped” into “engine repeatedly killed” — strictly worse than the failure it contains. Increment 3 measures it on GPU first, then ships the probe + condition together.

**Should-fix — `/dev/shm` misses source-level readOnly.** Correct, and the sharpest of the three: a source-level `readOnly: true` is **not** overridden by a mount-level `readOnly: false`, so my mount-level check plus a partial source list let NFS/CephFS/RBD/iSCSI/AzureFile/Quobyte/Portworx/ScaleIO/StorageOS through into exactly the silent no-transfer the guard exists to prevent. Now enumerates every in-tree source carrying the flag (plus AzureDisk, Gluster, Cinder, FlexVolume, GCE PD, EBS, FC), with tests for the representative ones — and a test for the inverse, that a **writable** NFS `/dev/shm` is still reused rather than over-rejected.

**Should-fix — VERSIONS.md tuple.** Correct, and now fillable: the row was a placeholder with a TODO awaiting a successful GPU build, and that build has since happened. Recorded the validated tuple — `(lmsysorg/sglang:nightly-dev-cu13-20260711-7de33ce8, lmcache 0.5.1)` — with the two constraints it encodes: **cu13 is load-bearing** (0.5.1 links `libcudart.so.13`), and the stock `v0.5.1.post2-cu126` tag **predates `--enable-lmcache`**. Also noted that the validation `pip install`ed lmcache at pod start, so a real deployment must bake it into a derived image and pin that digest — a runtime install is not a reproducible pin.

**Design-doc status.** The `sglang-lmcache-mp-mode.md` link now resolves — #146 merged, and this branch is rebased onto it. Its status text is updated in this commit to say Phase 2 landed (increments 1–2) rather than “implementation pending”.

EdHasNoLife:
Deferral tracked: **[CAC-227](https://linear.app/cachebox/issue/CAC-227/sglang-mp-finish-the-worker-spof-containment-engine-liveness-probe)** (High) — the engine liveness probe + `CacheBackend` `Degraded` on worker unhealth, with the mid-flight-worker-restart measurement as its first, blocking step. The design doc's containment table and phased-delivery section now carry the same split, so the promise is recorded in-repo rather than only in review comments.

EdHasNoLife:
Round 12 — all 10 doc findings fixed in 40341d1, and they were fair: my earlier edits updated specific rows while leaving the surrounding Phase-1 narrative describing the old broken `lm://` world, so both documents contradicted themselves. That is worse than either state alone — a reader cannot tell which half is current.

**Design doc** (`sglang-lmcache-mp-mode.md`):
- Status header + "warning already ships" → implemented, GPU-validated, warning retired.
- **Topology (`:179`)** — good catch, and the doc was the stale one, not the code: it specified a separate `lmcache-config` init container writing `/config/lmcache.yaml`, while the worker writes `/etc/lmcache/config.yaml` itself and then `exec`s the server. Synced to as-built, with why it is *fewer* moving parts: one process renders both sides of `mp_port`, and the `startupProbe` already gates the engine on the server listening — which implies the file exists. A separate init container was a second place to keep in sync for no added ordering guarantee.
- **Open question (`:317`)** — resolved, and the answer is more interesting than either branch it anticipated. The data path **is** CUDA-IPC (so the "if `non_gpu` works" branch is out), but its predicted consequence — *"a separate sidecar cannot get the engine's GPU, so fall back to one container"* — is **false**: a sidecar needs GPU **visibility**, not an **allocation**. So the clean design survived and the single-container fallback was never needed. The section now records that, the validated negative (`Device UUID … not found in the discovered devices`), and the isolation cost it locks in.
- TL;DR `:42` no longer calls sidecar packaging open.

**VERSIONS.md** — the sharpest finding here was `:23`, and it was **my own claim that validation disproved**: the row predicted the MP fix would reuse the `lm://` server behind a per-node worker. It cannot — `lm://` is not a valid `--l2-adapter` type. Prediction retired rather than left to mislead; the row stays, scoped to vLLM where it is correct. Also: the opening KNOWN LIMITATION (`:6`) now describes what SGLang actually pins; the pinning instructions (`:45`, `:60`) no longer tell operators to pin a server the wire never dials, and name the real constraints (**engine ↔ worker lmcache parity**, **lmcache ↔ CUDA runtime**) instead of the fictional lmcache ↔ lmcache-server one; and the reproducibility preamble (`:85`, `:107`) no longer says "no validated tuple exists" three lines above the tuple.

The reference **manifest** still carries the `example.invalid/sglang-lmcache` placeholder — that is deliberate and now stated as such: updating it belongs to the reference-stack increment, and the controller-rendered managed path does not read it.

EdHasNoLife:
Round 13 — Blocking + both should-fixes addressed in 9b55c2c (rebased onto the updated #147).

**Blocking — worker lacks a Restricted-compatible securityContext.** Real, and the mirror image of the earlier IPC_LOCK finding: `allowPrivilegeEscalation: false` and `capabilities.drop: [ALL]` are **container-only** Restricted requirements (not inheritable from the pod), so a worker without them turns an admissible engine pod into a rejected one in a Restricted namespace — the cache plane breaking the engine. Fixed:
- Always set `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `seccompProfile: RuntimeDefault`.
- Deliberately does **not** force `runAsUser` / `runAsNonRoot` / `readOnlyRootFilesystem` the way the distroless subscriber sidecar does — the worker runs the operator's engine image, whose UID and writable paths this adapter must not override (a forced UID or read-only rootfs can break CUDA-IPC or the image's own writes). Instead it **mirrors the engine container's user identity** (`runAsNonRoot`/`runAsUser`/`runAsGroup`) when set — same image by default, so the same user is valid — and inherits from the pod otherwise. Tests cover both the restricted set and the mirroring.

⚠️ **Validation status:** the `drop: [ALL]` change is **not yet GPU-re-validated** — every prior worker run carried the default Docker cap set, not drop-ALL. The reasoning is that CUDA-IPC uses device files + `/dev/shm`, not Linux capabilities (and the kvevent-subscriber already runs drop-ALL in engine pods), so it should hold — but I will confirm store→flush→retrieve still caches under this securityContext on the next cluster-token refresh (the OCI token expired again mid-session). Flagging rather than asserting.

**Should-fix — K8s ≥1.29 native-sidecar floor not prominent.** Right — it was only in the design doc. Added a prominent operator-facing caveat at the top of the SGLang support section in `cachebackend-api.md`: the MP wire has a hard 1.29+ cluster floor (native sidecars), and on older clusters a SGLang pod fails admission rather than failing open. No in-webhook version gate today; documented as an operator prerequisite.

**Should-fix — recipe still had a placeholder lmcache version.** Fixed: the derived-image recipe now pins `lmcache==0.5.1`, matching the GPU-validated tuple recorded in the row above it.

EdHasNoLife:
Round 14 — Blocking cleared (the securityContext fix landed); the 3 remaining were all doc-consistency, fixed in 69397b4. Two were consequences of changes earlier in this PR, so fair catches:

- **failOpen invariant contradicts the SGLang exception** (`cachebackend-api.md:470`) — right: the normative "MUST fall back to local prefill … never a serving dependency" line predated the pair-specific exception I added to the `failOpen` row. Qualified it: the invariant holds for a **remote/shared** tier; the co-scheduled SGLang MP worker is the one documented exception (fail-open still honored at the shared Redis L2).
- **Singleton claim vs. vLLM scale-out** (`sglang-lmcache-mp-mode.md:161`) — right, and it contradicted code I wrote in this PR: the doc said vLLM's `lm://` server "already carries the same constraint", but `rejectSGLangRedisL2ScaleOut` is deliberately scoped to `(sglang, LMCache)` and vLLM+LMCache scales/autoscales. Reworded to state the singleton is SGLang-Redis-L2-specific and to name both enforcement points.
- **"config-file init container" heading** (`:177`) — no such container exists; the native worker writes its own config. Renamed the heading + the two matching mentions.

No code change this round. #147 (the base) is fully Codex-clean.

EdHasNoLife:
Round 15 — Blocking none; the readiness Should-fix was a real bug, fixed in bdd0d08.

- **Singleton clamp not reflected in readiness** (`controller:1038`) — genuine bug, and it **pre-existed for the host-network Mooncake master**: `desiredReplicas` (the readiness expectation) returned the CR's `spec.replicas` (e.g. 3) while `clampSingletonReplicas` deployed one pod, so a grandfathered scaled-out singleton reports `RolloutInProgress` forever. The clamp generalization extended the latent bug to the SGLang Redis L2, and Codex catching it here is what surfaced it for Mooncake too. Fixed `desiredReplicas` to apply the same clamp (0 preserved), with tests for both singleton reasons, disabled-stays-0, and that vLLM+LMCache (not a singleton) still honors `spec.replicas`.
- **VERSIONS.md vs. the reference manifest** (`:43`) — right: the section describes the MP topology, but `manifests/sglang-lmcache/deployment.yaml` still ships the old lmcache-server placeholder (its rewrite is the increment-3 reference-stack leg). Added a note that the pins describe the **controller-rendered managed path** (authoritative, GPU-validated) and that the hand-written manifest lags until increment 3.
- **Nits** — both fixed: `applyDeployment` godoc now names `clampSingletonReplicas` + both singleton reasons (not the removed `clampHostNetworkToSingleton`); the sanitization comment now names `sglangIntInRangeOr` (was a nonexistent `sglangPositiveIntOr`).

EdHasNoLife:
Round 16 — Blocking none; both items were doc-consistency (the plan/background lagging what this PR landed), fixed in 08c11df.

- **Phased-delivery plan** (`:437`) — right: it still listed the `cachebackend-api.md` MP rewrite and the smoke warning-absence flip under Phase 3, but both landed in increment 2. Moved them out; Phase 3 now lists only the genuinely-remaining operator surface (samples polish, reference-stack manifest leg, and the smoke assertion that the wire is actually *injected*).
- **Background "current adapter"** (`:89`) — reworded to "the previous `lm://` adapter injected", since this PR drops that env.

No code change. #147 (base) remains fully Codex-clean.

heymrbox (approved):

PR #153, opened 2026-07-17
Add release SBOM generation

heymrbox:
## Summary

- Add a release SBOM workflow that installs Syft, builds the controller/server/kvevent-subscriber images, generates SPDX JSON SBOMs, and attaches them to GitHub Release artifacts.
- Add `make sbom`, `make sbom-release`, and `make sbom-images` so SBOM generation is reproducible locally and in CI.

## Linked issues

CAC-212

### Quality
- [ ] New/changed behavior has unit tests.
- [x] CI is green.


heymrbox:
Addressed in 719d001.

- Moved Syft installation into a shared `.github/actions/setup-syft` action and fixed checksum verification to run from `$RUNNER_TEMP`, where the archive is downloaded.
- Added a PR-time `SBOM Smoke` CI job that uses the same installer and runs `make sbom-release`, so the release-only path is exercised before merge.
- Updated the release workflow to generate image SBOMs from release-tagged `registry:` image references after GHCR login, rather than scanning freshly rebuilt runner-local images.
- Fixed the Makefile dependency graph so default `make sbom` is safe under `make -j`: `sbom-images` owns the image-build dependency, while registry mode skips the local build.
- Removed the unused Buildx setup; `packages: read` is now used by the release workflow for GHCR registry reads.

Verification:
- `make -n sbom TAG=v0.0.0-test`
- `make -n sbom-release sbom-images TAG=v0.0.0-test SBOM_IMAGE_SOURCE=registry`
- YAML parse for `ci.yml`, `release-sbom.yml`, and `setup-syft/action.yml`
- `make verify-naming`
- `make verify-no-internal-refs`
- `make lint`
- Syft checksum smoke: checksum verified OK locally; Linux binary execution is left to ubuntu CI
- `make pre-pr`

heymrbox:
Addressed the follow-up in e1c2660.

- Syft setup now verifies the downloaded linux-amd64 archive against a repo-pinned SHA-256 digest instead of downloading a mutable checksum file from the same release.
- The release SBOM workflow is self-contained again: it builds the three release-tagged images locally and scans those built images, so it no longer depends on an external GHCR publishing process.
- CI now validates both SBOM paths: the lightweight SBOM smoke covers `sbom-release`, and the image-build job runs `sbom-images` against the freshly built controller, server, and subscriber images.
- `sbom-images` now fails fast if Syft is unavailable before doing image work, while still allowing CI to skip duplicate image builds with `SBOM_IMAGE_BUILD=0`.
- Removed the now-unneeded `packages: read` permission from the release workflow.

Verification:
- `make -n sbom TAG=v0.0.0-test`
- `make -n sbom-images TAG=ci-test SBOM_IMAGE_BUILD=0 SBOM_DIR=/tmp/sbom-image-smoke`
- YAML parse for `ci.yml`, `release-sbom.yml`, and `setup-syft/action.yml`
- `git diff --check`
- repo-pinned Syft archive checksum verified locally
- `make verify-naming`
- `make verify-no-internal-refs`
- `make lint`
- `make pre-pr`

heymrbox:
Addressed the latest SBOM review in 1ebbc85.

- Added a release-tag guard before checkout/SBOM generation so release tags must be Docker-tag compatible before they flow into Make or image refs.
- Kept the SBOM Make recipes on the pinned Syft version by default and fail fast when the local Syft binary is missing or has the wrong version.
- Quoted the SBOM/image recipe arguments and Docker image tags.
- Tightened CI SBOM smoke checks so source and image SBOMs must have a non-empty packages array, not just valid JSON shape.

Verification:
- `make -n sbom TAG=v0.0.0-test`
- `make -n sbom-images TAG=ci-test SBOM_IMAGE_BUILD=0 SBOM_DIR=/tmp/sbom-image-smoke`
- YAML parse for `ci.yml`, `release-sbom.yml`, and `setup-syft/action.yml`
- `git diff --check`
- Release-tag regex positive/negative smoke
- Fake-Syft Makefile smoke for pinned-version check + non-empty SPDX assertion
- `make verify-naming`
- `make verify-no-internal-refs`
- `make lint`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`

heymrbox:
Addressed the follow-up in 248927d.

- The release workflow now checks out current default-branch SBOM tooling first, so the local setup action and Make SBOM targets are available even when backfilling an older existing release tag.
- The requested release tag is checked out separately under `release-source`.
- SBOM generation runs from `release-source` with the current Makefile via `make -f`, writing artifacts back to `$GITHUB_WORKSPACE/dist/sbom`.

Verification:
- `make -f "$PWD/Makefile" -C "$PWD" -n sbom TAG=v0.0.0-test SBOM_DIR=/tmp/ic-release-sbom`
- YAML parse for `release-sbom.yml`
- `git diff --check`

heymrbox:
Addressed the latest release-SBOM review in 4a51a9e.

- Manual dispatch now validates the GitHub Release up front and checks out `refs/tags/$RELEASE_TAG`, so a branch with the same name cannot be used as release source.
- Release SBOM tooling is checked out at the immutable workflow SHA with checkout credentials disabled.
- The release workflow now builds and pushes the controller, server, and subscriber release images, captures each pushed digest, and generates image SBOMs from `registry:...@digest` instead of runner-local tags.
- Upload is split into a separate contents-write job; the generation job keeps only read/package permissions.
- `make sbom-images` keeps recursive image builds on the intended Makefile, even when the source checkout lives in another directory.
- CI now uses one workflow-level Syft version pin for both SBOM smoke paths.

Verification:
- YAML parse for `ci.yml` and `release-sbom.yml`
- `make -n sbom-images TAG=v0.0.0-test`
- `make -f "$PWD/Makefile" -C "$PWD" -n sbom-release TAG=v0.0.0-test SBOM_DIR=/tmp/ic-release-sbom`
- `git diff --check`
- Pinned action SHA lookups for `docker/login-action`, `actions/download-artifact`, and `docker/setup-buildx-action`
- `make verify-naming`
- `make verify-no-internal-refs`
- `make lint`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`

heymrbox:
Addressed the canonical-tag overwrite concern in 34d4a5b.

- The release workflow no longer builds or pushes release image tags. It requires the release images to already exist, resolves each published tag to its registry digest, and scans `registry:repo@sha256...` for the image SBOMs.
- Added `make sbom-registry-images` as the shared/testable registry-digest scan path used by the release workflow.
- Added a PR-time fake-registry smoke in CI for that exact Make target, covering digest parsing, registry-style Syft refs, and non-empty SPDX output for controller/server/subscriber.
- Kept release-generation permissions at `contents: read` + `packages: read`; only the final attach job has `contents: write`.

Verification:
- YAML parse for `ci.yml` and `release-sbom.yml`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom`
- Local fake `docker buildx imagetools inspect` + fake Syft smoke for `make sbom-registry-images`
- `git diff --check`
- `make verify-naming`
- `make verify-no-internal-refs`
- `make lint`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`

heymrbox:
Addressed the missing-publisher / no-overwrite concerns in 1f0c325.

- `make sbom-registry-images` now resolves existing GHCR release tags first. If a digest already exists, it scans that immutable `registry:repo@digest` ref and does not push anything.
- The release workflow opts into `SBOM_REGISTRY_PUBLISH_MISSING=1`, so it only builds and pushes a release image when that tag is missing. That gives normal releases a self-contained path without replacing already-published images or manifest lists.
- The registry target now uses `DOCKER_BUILD_CMD` and configurable build context/Dockerfile values.
- CI’s registry-path smoke now exercises the missing-tag publish branch with stricter fakes: Docker validates expected image refs/tags and Syft validates the immutable registry digest refs.
- The Syft setup action now fails if the version is overridden without also supplying the matching archive checksum.

Verification:
- YAML parse for `ci.yml`, `release-sbom.yml`, and `setup-syft/action.yml`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- Strict local fake `docker buildx imagetools inspect` / `docker buildx build` / Syft smoke for `make sbom-registry-images`
- `git diff --check`
- `make verify-naming`
- `make verify-no-internal-refs`
- `make lint`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`

heymrbox:
Addressed the registry inspect failure mode in 12750a1.

- `sbom-registry-images` now captures `imagetools inspect` stderr and only treats known missing-manifest errors as publishable.
- Auth, registry, Docker, Buildx, and other operational failures now fail closed and do not trigger a build/push.
- The CI registry-path smoke now has two cases: an auth failure must stop without publishing, while a confirmed missing manifest may build, read pushed metadata, and scan the immutable registry digest ref.

Verification:
- YAML parse for `ci.yml`, `release-sbom.yml`, and `setup-syft/action.yml`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- Local fake registry smoke covering auth failure and confirmed missing-manifest publish path
- `git diff --check`
- `make verify-naming`
- `make verify-no-internal-refs`
- `make lint`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`

heymrbox:
Addressed in 1609abe.

- Tightened `sbom-registry-images` so only explicit manifest/name-missing registry responses can enter the publish-missing path; broad `not found` errors now fail closed.
- Added CI smoke coverage for a credential-helper failure, so that shape cannot be mistaken for a missing release image.
- Added the normal existing-image path to the registry SBOM smoke with real-ish `imagetools inspect` output, covering digest parsing and registry `repo@digest` Syft refs without building.

Verification:
- YAML parse for `.github/workflows/ci.yml`
- `git diff --check`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- Local fake-registry smoke for auth failure, credential-helper failure, existing-image digest parsing, and confirmed missing-manifest publish
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make ci`

heymrbox:
Addressed in a27b675.

- Added `SBOM_IMAGE_PLATFORMS` with the release fallback defaulting to `linux/amd64,linux/arm64`.
- Threaded that platform set into the publish-missing `docker buildx build --push` path so the SBOM workflow cannot silently publish a runner-default single-arch image.
- Extended the CI fake-registry smoke to parse and assert the `--platform linux/amd64,linux/arm64` flag on the missing-image build path.

Verification:
- YAML parse for `.github/workflows/ci.yml`
- `git diff --check`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- Local fake-registry smoke with platform validation
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make ci`

heymrbox:
Addressed in a6ac498.

- `sbom-registry-images` now scans each configured release platform separately when `SBOM_IMAGE_PLATFORMS` is set.
- The default release platform set remains `linux/amd64,linux/arm64`, so published multi-platform image indexes produce platform-specific image SBOMs instead of a runner-platform-only SBOM.
- Output names now include the platform suffix, e.g. `inference-cache-controller-linux_amd64-<tag>.spdx.json` and `...-linux_arm64-<tag>.spdx.json`.
- CI fake-registry smoke now requires both platform scans for existing-image and missing-image paths.

Verification:
- YAML parse for `.github/workflows/ci.yml`
- `git diff --check`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- Local fake-registry smoke covering auth failure, credential-helper failure, existing-image per-platform scans, and missing-image per-platform scans
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make ci`

heymrbox:
Addressed the nits in 11a4338.

I moved the registry-image SBOM smoke out of the workflow into `hack/sbom-registry-smoke.sh`, so the CI step is now a small call into a reusable script and the same hardening path can be run locally.

I also added `make verify-syft-pin`, wired it into CI and the local `make ci` gate, and left a note on the setup action default so Syft version/checksum bumps fail fast if the Makefile, workflows, and action drift apart.

Verification:
- `bash -n hack/sbom-registry-smoke.sh hack/verify-syft-version.sh`
- `make verify-syft-pin`
- YAML parse for CI/release/action files
- `git diff --check`
- `IMAGE_TAG=ci-local bash hack/sbom-registry-smoke.sh`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`

heymrbox:
Addressed in 265a373.

I fixed the registry SBOM platform handling so existing images are scanned only for platforms that `docker buildx imagetools inspect` actually reports. If an existing image has no manifest platform list, the target falls back to a single digest scan without forcing a platform; the publish-missing path still scans the configured platforms because it just published that set.

I also extended `hack/sbom-registry-smoke.sh` with an amd64-only existing-image case, tightened `hack/verify-syft-version.sh` to parse effective fields instead of substring-matching, and added SBOM usage docs at `docs/operations/sbom.md` with a README link.

Verification:
- `bash -n hack/sbom-registry-smoke.sh hack/verify-syft-version.sh`
- `make verify-syft-pin`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- `git diff --check`
- `IMAGE_TAG=ci-local bash hack/sbom-registry-smoke.sh`
- `HOME=/tmp/ic-home GOPATH=/tmp/ic-go GOCACHE=/tmp/ic-go-build-cache GOMODCACHE=/tmp/ic-go-mod-cache GOLANGCI_LINT_CACHE=/tmp/ic-golangci-lint-cache PYTHONPATH=/tmp/ic-python-deps make pre-pr`

heymrbox:
Addressed the final doc nits in 45f63ff.

- Added the aggregate `make sbom` target to the README common-targets list.
- Clarified the local Syft install guidance to point at the Makefile `SYFT_VERSION` value instead of implying `make syft-check` prints the version.

Verification:
- `git diff --check`
- `make verify-no-internal-refs verify-syft-pin`

EdHasNoLife (approved):

Makefile
```diff
@@ -311,15 +326,133 @@ image-build: controller-image server-image subscriber-image ## Build controller,
...
+		if $(DOCKER_BUILD_CMD) buildx imagetools inspect "$$ref" >"$$inspect_log" 2>&1; then \
+			digest="$$(awk '/^Digest:/ {print $$2; exit}' "$$inspect_log")"; \
+			if [ -z "$$digest" ]; then \
+				echo "ERROR: unable to resolve digest for $$ref" >&2; \
+				cat "$$inspect_log" >&2; \
+				rm -f "$$inspect_log"; \
+				exit 1; \
+			fi; \
+			available_platforms="$$(sed -n 's/^[[:space:]]*Platform:[[:space:]]*//p' "$$inspect_log" | sort -u | tr '\n' ',' | sed 's/,$$//')"; \
+		elif grep -Eiq 'manifest unknown|manifest not found|no such manifest|name unknown' "$$inspect_log"; then \
```
EdHasNoLife: The `sbom-registry-images` target treats an image as "missing → build+push" **only** when `docker buildx imagetools inspect` stderr matches `manifest unknown|manifest not found|no such manifest|name unknown`; anything else falls through to the `else` branch and `exit 1`.

This path is load-bearing: no other workflow or `make` target pushes the release images to ghcr (the only `--push` in the repo is line 407 here), so on release `release-sbom.yml` runs this with `SBOM_REGISTRY_PUBLISH_MISSING=1` as the sole publish-and-SBOM mechanism. On the **first** release the `inference-cache-{controller,server,subscriber}` packages don't exist yet, and ghcr's "does-not-exist" response for a brand-new package can surface differently from a plain missing-tag (`name unknown`, `denied`, or `<ref>: not found`) — any wording not in the regex fails the whole release job instead of publishing.

The direction is safe (fail-closed, visible failure — not a silently-skipped SBOM), and the bundled smoke can't catch it because the fake `docker` returns exactly `manifest unknown`. **Suggest** confirming the real `imagetools inspect` output against a missing ghcr tag/package before the first tagged release, and widening the regex to cover whatever ghcr actually emits for a not-yet-published package. Keep the auth/credential errors (authfail/helperfail) routing to the hard-error branch as they do today — that distinction is correct and tested.

(Flagged from inspection + reasoning about the release path; I could not reproduce ghcr's exact `imagetools inspect` error here.)

.github/actions/setup-syft/action.yml
```diff
@@ -0,0 +1,39 @@
...
+    run: |
+      set -euo pipefail
+      default_version="v1.20.0"
+      default_sha256="689e12c5cbf67521ce61b9c126068f9eaabe1223e77971b2fede50033ff6b5cc"
+      if [ "$SYFT_VERSION" != "$default_version" ] && [ "$SYFT_SHA256" = "$default_sha256" ]; then
+        echo "linux-amd64-sha256 must be supplied when overriding Syft version" >&2
+        exit 1
+      fi
+      version="${SYFT_VERSION#v}"
+      archive="syft_${version}_linux_amd64.tar.gz"
```
EdHasNoLife, line 40: `archive="syft_${version}_linux_amd64.tar.gz"` plus a single `linux-amd64-sha256` input means the action silently assumes a linux/amd64 runner. Correct for every current caller (all `ubuntu-latest`), so this is a nit — but the composite action has no guard rejecting a non-amd64 host, so a future reuse on an arm64/macOS runner would download the amd64 tarball (checksum mismatch at best, wrong-arch binary at worst). Optional: add a `case "$(uname -sm)"` guard that errors on anything other than the supported arch, or derive the arch into the archive name + checksum.

EdHasNoLife (commented):
## Review: Add release SBOM generation (CAC-212)

Reviewed the SBOM tooling end-to-end — the pinned Syft install action, the `sbom*` Makefile targets, the `sbom-smoke` CI job + image-build SBOM step, and the `release-sbom.yml` generate/attach workflow. This delivers the ticket DoD (per-release + per-image SPDX SBOMs generated in CI and attached to the GitHub Release). Nicely self-tested and hardened.

**What I actually ran (against the PR head, `45f63ff`):**
- `make verify-syft-pin` → green; the Syft pin is synchronized across the Makefile, both workflows, and the setup action.
- `bash hack/sbom-registry-smoke.sh` → passes. This drives the whole `sbom-registry-images` shell loop through fake `docker`/`syft`: fail-closed on auth/credential errors, multi-arch → 6 platform-suffixed SBOMs, single-arch → amd64-only (no arm64 file), and publish-missing → build+push+scan.
- Verified the pinned SHA-256 `689e12c5…6b5cc` **matches upstream** `syft_1.20.0_linux_amd64.tar.gz` exactly (fetched anchore's `syft_1.20.0_checksums.txt`); the release + archive naming are correct.
- `make -n sbom-release` / `make -n sbom-images` → target expansion correct; output filenames match the `Verify SBOM output` / `Verify image SBOM outputs` CI checks.
- `bash -n` on both new scripts → clean; `make verify-naming` + `make verify-no-internal-refs` → clean; no `oci`/`oracle` (the only hits are the OCI *Open Container Initiative* media type and the verify-naming target's own pattern) and no assistant leakage in the changed files.

**Findings** (both non-blocking):
- **Should verify before first release** — `Makefile:394`: the publish-missing fallback keys on a hardcoded set of registry error strings. Since `release-sbom.yml` is the *only* path that pushes release images to ghcr (sole `--push`, line 407), this is load-bearing on release; a ghcr "package/tag does not exist" response that doesn't match the regex fails the release job instead of publishing. Fail-closed and visible, and the smoke can't catch it (fake returns exactly `manifest unknown`). Worth confirming ghcr's real `imagetools inspect` output for a missing package before the first tagged release.
- **Nit** — `.github/actions/setup-syft/action.yml:33`: hardcoded `linux_amd64` archive + amd64-only checksum, no OS/arch guard. Correct for all current `ubuntu-latest` callers; only matters if the action is reused on another arch.

**Note, not a finding:** publishing the release images happens as a side effect of the SBOM workflow's publish-missing fallback — there's no separate release-images workflow today. Functional, but worth a conscious call on whether image publish should be its own step.

Verdict: approve-with-nits — solid, well-tested change; the one item above is a "confirm against real ghcr before the first release," not a code defect I could reproduce.

.github/actions/setup-syft/action.yml
heymrbox: Addressed in c032263.

Added an explicit `Linux x86_64` guard before the action downloads the linux-amd64 Syft archive. Current callers still pass unchanged, but future non-amd64 reuse now fails with a clear message instead of downloading the wrong artifact.

Verification:
- YAML parse for `.github/actions/setup-syft/action.yml`
- `make verify-syft-pin`
- `make pre-pr`

Makefile
heymrbox: Addressed in c032263.

I kept explicit auth/credential failures on the hard-error path, then widened the missing-image matcher to cover GHCR brand-new package/tag shapes like `denied: requested access to the resource is denied` and `<ref>: not found`. I also extended `hack/sbom-registry-smoke.sh` so the fake registry covers those denied/not-found missing paths in addition to the existing manifest-unknown path.

Verification:
- `IMAGE_TAG=ci-local bash hack/sbom-registry-smoke.sh`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- `make pre-pr`

heymrbox:
Addressed the SBOM registry denial review in 7c7ae4d.

`denied: requested access to the resource is denied` now stays on the hard-error path, so release automation won't build or push when registry access is misconfigured. The publish-missing fallback still covers manifest-missing and `<ref>: not found` responses.

Verification:
- `bash -n hack/sbom-registry-smoke.sh hack/verify-syft-version.sh`
- `make verify-syft-pin`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- `git diff --check`
- `IMAGE_TAG=ci-local bash hack/sbom-registry-smoke.sh`
- `make pre-pr`

heymrbox:
Addressed the single-manifest SBOM smoke coverage review in 9af4b52.

The registry smoke now has an `existing-single` inspect mode with a digest and no `Platform:` rows, and it verifies the Makefile writes the unsuffixed SBOM file without invoking Syft with `--platform`.

Verification:
- `bash -n hack/sbom-registry-smoke.sh hack/verify-syft-version.sh`
- `make verify-syft-pin`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- `git diff --check`
- `IMAGE_TAG=ci-local bash hack/sbom-registry-smoke.sh`
- `make pre-pr`

heymrbox:
Addressed the SBOM publish-fallback smoke gap in f5c253d.

The fake `docker buildx build` path now verifies `--push`, the requested platform list, Dockerfile, component target, tag, metadata file, and build context. It derives the expected target from the image tag, so the smoke will fail if the release path stops publishing or builds the wrong component/source.

Verification:
- `bash -n hack/sbom-registry-smoke.sh hack/verify-syft-version.sh`
- `make verify-syft-pin`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- `git diff --check`
- `IMAGE_TAG=ci-local bash hack/sbom-registry-smoke.sh`
- `make pre-pr`

heymrbox:
Addressed the registry metadata parsing review in 618ca54.

`sbom-registry-images` now calls `docker buildx imagetools inspect --format {{json .Manifest}}` and parses digest/platforms with `jq` instead of scraping human-readable output. The smoke fake now requires the structured inspect call and emits matching manifest JSON fixtures. I also removed the one-item `missing-notfound` loop.

Verification:
- `bash -n hack/sbom-registry-smoke.sh hack/verify-syft-version.sh`
- `make verify-syft-pin`
- `make -n sbom-registry-images TAG=v0.0.0-test SBOM_DIR=/tmp/ic-registry-sbom SBOM_REGISTRY_PUBLISH_MISSING=1 SBOM_IMAGE_CONTEXT=/tmp/source SBOM_DOCKERFILE=/tmp/source/dockerfiles/Dockerfile`
- `git diff --check`
- `IMAGE_TAG=ci-local bash hack/sbom-registry-smoke.sh`
- `make pre-pr`

PR #152, opened 2026-07-16
feat(lookup): carry T1/T2/T3 tier in LookupRoute replica scores

EdHasNoLife:
Tier-aware LookupRoute — schema + plumbing only (CAC-172). Tier *detection* (T2 offload vs T1) is a separate ticket and is intentionally out of scope.

**Proto (additive, v1alpha1-compatible)**
- New `CacheTier` enum: `CACHE_TIER_UNSPECIFIED | _T1 | _T2 | _T3`.
- New `ReplicaScore.tier` (field 5) — the tier from which the replica can serve the **entire** matched prefix: the **least-local (coldest) tier across the matched run**. A T1+T2 run reports T2 — the constraining block defines the tier (per review round 1).

**Index**
- Tier tag on index entries (`replicaEntry.tier`, `PrefixRef.Tier`, `ReplicaScore.Tier`).
- Default-ingest rule: an entry with no tier is tagged `T1` (engine KV cache). Per-block/entry tier remains most-local as reported at ingest.
- Chain match folds across the run with `worstTier` (T3 ⊃ T2 ⊃ T1); an `UNSPECIFIED` block tier poisons the fold to `UNSPECIFIED` rather than fabricating a claim. Ranker scoring unchanged — tier is carried, not yet an input.

**Server**
- Map the index tier onto the wire `ReplicaScore.tier`; non-prefix hints (TENANT_HOT / AFFINITY_HINT) stay `UNSPECIFIED`.

**Generated / docs**
- Regenerated protoc output (`make proto-gen`, protoc-gen-go v1.36.5); `docs/design/grpc-contract.md` documents the coldest-tier semantics.

**Tests**
- Index: tier defaults to T1 on ingest (exact + chain), explicit tier round-trips, coldest-tier fold across a mixed-tier run (T1+T2 ⇒ T2), `worstTier` unit table incl. UNSPECIFIED-poisons.
- Server: tier round-trips index → LookupRoute response (T1); TENANT_HOT leaves tier UNSPECIFIED; full `cacheTierToProto` mapping.

**Validation**
- `make ci` (fmt + vet + golangci-lint + prometheus + golden vectors + race tests + build): pass
- `make proto-lint`, `buf breaking` vs main, `make verify-naming`, `make verify-no-internal-refs`: pass
- `make cover-check`: 91.8% (min 90%)

PR #154, opened 2026-07-17
feat(subscriber): tag cache tier T1/T2 from block lifecycle

EdHasNoLife:
Implements cache-tier (T1/T2) detection + tagging in the kvevent-subscriber. Ref CAC-173.

## Design decision (the hard part: how to detect T2)

vLLM's KV-event channel (`pkg/adapters/engine/events.go`) emits exactly `BlockStored` / `BlockRemoved` / `AllBlocksCleared`. **It announces the T1 (HBM) lifecycle but emits *nothing* when LMCache offloads a block to L2** — the `LMCacheConnectorV1` offload is invisible to the KV-event surface. So T2 is **not directly observable**; it must be *inferred*.

Resolved model (from the signals that actually exist):

- `BlockStored` → **T1** (resident in HBM).
- `BlockRemoved`, replica has an L2 tier → **T2 downgrade**: the block left HBM but LMCache still holds it, so re-report the same content-fingerprint key at T2 (reload-able) via the additive `ReportCacheState` path instead of deleting it. The index applies last-write-wins on tier, moving the entry T1→T2. A later `BlockStored` of the same content re-reports it at T1 (upgrade).
- `BlockRemoved`, no L2 tier → **delete** (`PREFIX_EVICTED`), as before.
- **T3** (remote) is out of scope — no engine signal distinguishes it today.

**Honesty caveat (documented in the design note):** a T2 tag is a *prediction of reload-ability*, not confirmation the block is actually in LMCache. An offload can silently fail; routing stays fail-open, so a wrong T2 hint just recomputes (a cache miss, never a wrong answer). Detecting a silently-degraded offload tier is a separate health concern (the `t2_hit_tokens`/`t2_query_tokens` counters on `ReplicaStats`), not this ticket.

This is the ticket's "combine cold observation (`BlockRemoved`) with warm prediction (L2 retains it)" goal: keep a *usable, honestly-tagged* hint through the HBM→L2 transition instead of going blind (previously this path silently kept the entry stale at **T1**, over-claiming HBM residency).

### How the subscriber knows it has an L2 tier

Reused the existing `--ignore-block-removed` flag — it is already set **per integration mode** by the shared `RenderSubscriberSidecar` helper (Offload → set, EventsOnly → omit), precisely because the L2 tier is present only in Offload mode. Least-invasive, consistent with how the subscriber is launched today. The flag name predates the T2 downgrade and is kept for compatibility; its help text + the design doc are updated.

## token_ids re-hash — already present

Part 1 of the ticket (deterministic re-hash of `token_ids` → content-addressed index) **already landed** via the content-fingerprint work: `positional.go` derives the fingerprint from `token_ids` (`fingerprint.PrefixHashesFrom`); engine block hashes are used only as reverse-map keys and never forwarded. This PR is net-new **tier tagging** only.

## Changes

- `proto`: add `PrefixEntry.tier` (field 5, additive/v1alpha1-compatible); regenerated; `grpc-contract.md` updated in the same change.
- `pkg/server/inferencecache_service.go`: `updateFromProto` maps the wire tier → `index.PrefixRef.Tier` via new `cacheTierFromProto` (unset/unknown → T1 at ingest).
- `pkg/adapters/engine/positional.go`: stamp T1 on stored entries; `Removed` now returns the prefix hash **and** token count (needed to re-report at T2).
- `pkg/adapters/engine/forwarder.go`: L2-tier `BlockRemoved` re-reports at T2 through the pending/`ReportCacheState` path; single-tier path unchanged.
- `cmd/kvevent-subscriber/main.go`: updated `--ignore-block-removed` help.
- `docs/design/kvevent-subscriber-wiring.md`: rewritten "L2 cache tier semantics" section = the design note.
- The index tier carry/fold (`worstTier`, `normalizeIngestTier`, response plumbing from #152) is untouched.

## Tests

`BlockStored → T1`; `BlockRemoved`+L2 → entry retained as **T2** (no `PREFIX_EVICTED`); `BlockRemoved` w/o L2 → deleted; evict-then-rehit tier history `[T1, T2, T1]`; and tier surfacing end-to-end through `LookupRoute` (T1/T2/unset→T1). Updated `TestLMCacheOffloadKeepsRoutingHintWithIgnoreBlockRemoved` to assert the surviving hint is now tagged T2 (and to use a realistic eviction timestamp — the T2 re-report anchors freshness at the eviction time).

## Validation

`go build ./... && go vet ./... && go test ./...` pass; `make ci` green (pre-push gate); `make cover-check` 91.8% (floor 90%); `make proto-gen` idempotent (no drift).


EdHasNoLife:
Addressed the two SHOULD-FIX findings from the review (commit 8a03019):

- **T2 freshness timestamp overwrite (forwarder):** a `CacheStateUpdate` carries one `timestamp_us` for all its prefixes, so batching a T2 downgrade with a later `BlockStored` in the same debounce window let the store's timestamp refresh the T2 entry's freshness away from the eviction time. The downgrade is now flushed at the eviction boundary as its **own** update carrying the eviction timestamp (buffered adds flushed first to preserve store→evict order; `sendAdds` is synchronous). New test `TestReporterL2ModeDowngradeKeepsEvictionTimestamp` pins it.
- **Unknown future tier → T1 (server):** `cacheTierFromProto` now distinguishes `CACHE_TIER_UNSPECIFIED` (legacy-unset → T1 ingest-default) from an unrecognized non-zero value (a future colder tier), which is retained rather than collapsed to unset — so a future hold is never silently over-claimed as the hottest tier. Pinned by `TestCacheTierFromProtoDistinguishesUnsetFromUnknown` plus an end-to-end lookup case asserting an unknown tier surfaces `UNSPECIFIED`, not `T1`.

`go build/vet/test`, `make ci`, and `make cover-check` (91.9%, floor 90%) all green.

EdHasNoLife:
Final cleanup pass (commit f0ce6eb):

- **Nit fixed:** corrected the stale comment in `TestReporterL2ModeDowngradesEvictionToT2` — it claimed the T1 add and T2 downgrade "both sit in pending and flush together at shutdown." The reporter actually flushes the buffered T1 add first (preserving store→evict order) and sends the T2 downgrade in its own CSU anchored at the eviction timestamp, so the `[T1, T2]` wire history is deterministic. Comment-only; assertion unchanged.
- **Synchronous per-removal send (should-fix): intentionally not changed here.** Each L2 `BlockRemoved` does a `flush()` + synchronous send — but that's the same pattern the pre-existing single-tier removal path already uses (`flush()` + a `publish()` per evicted prefix). It's not a regression from this tier-tagging PR, and re-architecting the removal send-path for backpressure (which would affect both tiers) is out of scope here. Tracked as a separate backpressure-hardening follow-up (CAC-230).

`go build/vet/test`, `make ci`, and `make cover-check` (91.9%, floor 90%) all green.

EdHasNoLife:
Doc drift fixed (commit 0cd66da) — the last two references to the pre-change behavior:

- `docs/design/cachebackend-api.md`: the Offload eviction bullet no longer says `BlockRemoved` is suppressed + aged out on TTL; it now describes the T2 re-report (entry kept, tagged T2/reload-able, anchored at the eviction timestamp; re-store → T1), with EventsOnly/single-tier still evicting.
- `docs/reference/metrics.md`: `inferencecache_index_entries` falls on `BlockRemoved` only in single-tier mode; in L2/Offload mode the entry is retained via the T2 re-report (ages out via TTL), and tier changes on an existing `(scheme, hash)` don't move the gauge.

Docs-only; no behavior change. `go build/vet/test`, `make ci`, `make cover-check` (91.9%) all green.

heymrbox (approved):

EdHasNoLife:
Re: the Codex `forwarder.go` finding — L2/Offload mode re-reports an SGLang `BlockRemoved` as a T2 hint even though the SGLang integration doesn't offload yet (per `cachebackend-api.md`), producing a false T2 for SGLang replicas.

Tracked as **CAC-250** — deferred deliberately, not fixed here:
- **Out of scope** for this PR: #154 is the proto→index tier *mapping*; this is the eviction-forwarding policy (`RenderSubscriberSidecar` setting `--ignore-block-removed` for Offload adapters).
- **Degrades safely**: a false T2 → IC routes there → cache miss → served from scratch. Never a wrong answer, just a wasted routing decision (phantom entry ages out at TTL).
- **Latent**: only affects SGLang + LMCache *Offload* — a config that shouldn't be used until SGLang offload exists.

The fix (gate the flag on a functioning L2, or have the SGLang adapter opt out) should land *with* SGLang offload so the flag and the data plane stay consistent.

Ticket: https://linear.app/cachebox/issue/CAC-250

EdHasNoLife:
Update — resolved. The unblock landed: **#149 (SGLang LMCache MP-mode working data plane) is merged**, so SGLang now genuinely offloads to a managed Redis L2 (MP-worker sidecar + `--lmcache-config-file`), and the advisory "caches nothing" warning is gone. `--ignore-block-removed`→T2 on a SGLang Offload backend now has a real L2 behind it — the hint is **correct**, not false.

**CAC-250 closed as resolved-by-#149.** The only residual (a *transient* L2/worker outage tagging T2 while the block isn't actually there) is general soft-state — safe — and lives in CAC-227 (worker-SPOF containment), not here.

PR #160, opened 2026-07-22
feat(subscriber): read engine load via GetLoads gRPC (CAC-235)

EdHasNoLife:
Gives the kvevent-subscriber a **gRPC-native load source**: read the engine's `GetLoads` RPC and project `SchedulerLoad` into `ReplicaStats`, as an alternative to scraping HTTP `/metrics`.

**Why:** a vLLM engine served over gRPC (`vllm.entrypoints.grpc_server`) exposes **no HTTP `/metrics`** — live load is available only over `GetLoads`. Today the subscriber only scrapes HTTP, so IC gets no load and the ranker routes on residency alone. This is the clean long-term fix: structured data, no text round-trip, no extra container.

**Design:** `GRPCLoadsScraper` implements the existing `statsScraper` interface, so `StatsReporter` is unchanged. Selected by a new `--engine-loads-grpc <host:port>` flag (empty = HTTP scrape, unchanged default). Mapping mirrors the HTTP scraper: request counts to `pressure = clamp01((running+waiting)/ceiling)`, `token_usage` to `cache_memory_bytes` (via `--engine-cache-size-bytes`), `cache_hit_rate` to `hit_rate`; across DP ranks counts SUM, usage MEAN (capacity-weighted — each DP rank owns ~1/N of total capacity, so the average per-rank fill maps to true total bytes), hit-rate MEAN (finite values only).

**Vendored proto:** a minimal `GetLoads`-only subset of the **vLLM engine gRPC contract** (`pkg/adapters/engine/vllmengine/`), package kept `vllm.grpc.engine` for wire compat, deliberately **outside `proto/`** so it isn't subject to IC's buf-lint / drift checks (it's a vendored external contract, not IC's API). Regenerable via `make proto-gen-vendored`; CI drift-checks it.

**Caveat:** `GetLoads` carries no external-tier (T2/LMCache) token counters, so those `ReplicaStats` fields stay HTTP-/metrics-only in this mode; the load signals the ranker uses (pressure, cache-usage, hit-rate) are all provided.

Tests: mapping (single/multi-rank, zero-config, sanitization, overflow, error) + a bufconn end-to-end through the generated stubs + load-source control-flow selection. Full `pkg/adapters/engine` suite passes. Live end-to-end validation is pending a GetLoads-capable (vLLM >= 0.19) engine.

CAC-235.


EdHasNoLife:
Re: the Codex finding that `--engine-loads-grpc` isn't reachable through the auto-injected sidecar (`RenderSubscriberSidecar` leaves it on the HTTP `/metrics` default), so K8s-deployed gRPC engines still report no load.

Tracked as **CAC-249** — deferred deliberately, not wired here:
- **Premature vs. the engine floor**: `GetLoads` needs SMG servicer ≥ 0.5.2 / vLLM ≥ 0.19; the deployed engine is 0.18 / 0.5.1. Auto-wiring the flag now points every injected sidecar at an unimplemented RPC → each flips to the stale-load escalation from #156 fleet-wide — strictly worse than the HTTP default.
- **Needs a config surface**: selecting gRPC vs HTTP + the engine gRPC address is a new `CacheBackend.spec` field + renderer plumbing (a CRD API change, already a documented non-decision in `docs/design/kvevent-subscriber-wiring.md`).

Ships once the engine floor moves. The subscriber capability lands in this PR; the operator wiring follows in the ticket.

Ticket: https://linear.app/cachebox/issue/CAC-249

PR #162, opened 2026-07-23
docs(sglang): rewrite the reference stack to the MP data plane

EdHasNoLife:
The SGLang reference stack still described the **pre-MP `lm://` model** with a pervasive "KNOWN LIMITATION / caches nothing / do not use as a working offload / drop `--enable-lmcache`" narrative — now **false**: the MP data plane shipped (#149), so SGLang genuinely offloads. An operator following the reference was told a working feature is broken.

Rewrites the reference stack to the shipped **MP topology**:
- **`manifests/sglang-lmcache/deployment.yaml`** — replaces the standalone `lm://` `lmcache-server` + `LMCACHE_REMOTE_URL` env with a **Redis L2** store + a **node-local MP-worker native sidecar** (an `initContainers` entry with `restartPolicy: Always`) that writes the `--lmcache-config-file` and offloads to Redis via the `resp` `--l2-adapter`; the engine gets `--enable-lmcache` + `--lmcache-config-file`, and the lm:// env is dropped.
- **`manifests/sglang-lmcache/README.md`** — rewritten to the working MP flow (deploy Redis L2 + engine + worker, drive traffic, assert publisher-start + a non-empty Redis L2 after eviction), keeping the still-valid off-GPU wire tests, privacy boundary, managed-path, and block-size sections.
- **`VERSIONS.md`** — refreshes the note that flagged the manifest as lagging (it now matches). The pins themselves were already MP.

**Validation.** Derived from the **GPU-validated adapter render** (`redis_l2.go` + `internal/enginewire/sglang_mp.go`; the controller-rendered managed path was validated store→flush→retrieve in the MP increment) and **structurally validated** (`kubectl apply --dry-run=client` passes all four objects; native-sidecar `restartPolicy` recognized). It has **not** been independently re-run end-to-end on a GPU in this hand shape — the manifest + README + VERSIONS all say so and recommend a GPU run before treating it as golden.

**Still deferred under CAC-251** (not in this PR): the **MP smoke assertion**. It modifies the shared 3,600-line per-PR install-smoke gate, which I can't run locally — adding an unvalidated assertion there is high-blast-radius (a wrong resource name / timing budget breaks CI for the whole repo), so it should be added with a real kind run, not blind. (Also corrected while here: `cachebackend-api.md` and `VERSIONS.md`'s pins were already MP — my earlier "still open" list over-counted; see the CAC-251 comments.)

CAC-251.


EdHasNoLife:
Dispositioning the two remaining Codex findings (PR is otherwise green/mergeable):

- **Install-smoke assertion** — deferred to **CAC-251**, not added here. This PR is docs-only: it changes a **GPU-only reference manifest with non-applyable placeholder digests** the smoke gate can't apply (the gate installs `config/default` + `config/samples/`, not `docs/reference-stack/manifests/`). The MP-worker injection surface itself landed in **#149**; its managed-path reconcile smoke assertion is the tracked CAC-251 item. Adding it means editing the shared ~3,600-line per-PR install-smoke gate, which wants a real kind run to validate — deliberately not done blind.
- **Pin the exact publisher-startup log line** — can't be pinned without a GPU run of the specific SGLang image to observe the real phrasing; the "confirm against your build's first-run logs" caveat (pre-existing) is the honest handling.

CAC-251.

heymrbox (approved):

PR #168, opened 2026-07-28
docs(site): add Hugo documentation site + GitHub Pages workflow

EdHasNoLife:
Adds the Hugo/Docsy documentation site under `site/` (mirrors the OME docs layout) plus a GitHub Pages deploy workflow. Separate from the README refresh (#148).

**Contents**
- 33 content pages: overview, installation, quickstart, concepts (CacheBackend / CachePolicy / CacheTenant / CacheIndex / PromptTemplate / PDTopology + LookupRoute + gRPC contract), tasks, administration/operations, reference (CRD API, gRPC, metrics, reason codes, CLI doctor).
- Builds with Hugo Extended — no npm toolchain.
- `.github/workflows/docs.yml` deploys to GitHub Pages (free for public repos).

**To go live**
1. Merge this PR.
2. Enable Pages: Settings → Pages → Source: **GitHub Actions** (admin, one-time).

Serves at `https://cachebox-project.github.io/inference-cache/`.

Note: the Docs workflow deploy step fails until Pages is enabled — expected/harmless; it goes green on the first run after the toggle.


heymrbox (approved):

heymrbox:
Addressed in 9320d16.

The gRPC concept and reference pages now match the current proto: they include the adapter partition and cache-tier fields, document prompt-only token echoing, and call out the narrow LFU bookkeeping side effect. I also updated the LookupRoute concept page so it does not repeat the stale contract.

The docs workflow now builds on pull requests and deploys only for non-PR runs. It uses Hugo Extended directly without the redundant Dart Sass install, while the local setup now uses `npm install` for the optional PostCSS path.

Finally, the sitemap emits the configured absolute URLs, the unsupported CC BY footer claim is removed, and the unused Netlify redirects file and broken web manifest are gone.

Verification:
- Hugo Extended v0.164.0 production build
- generated sitemap URL inspection
- actionlint `.github/workflows/docs.yml`
- `make ci` via the pre-push hook

heymrbox:
Addressed the follow-up review in a15c1df.

All 113 internal documentation links now use Hugo `relref`, so the generated URLs retain the `/inference-cache/` GitHub Pages prefix. I also made the RSS, Prism stylesheet, and include-shortcode download links base-path-aware.

The installation guide now establishes a repository checkout, while the CPU quickstart uses a directly fetchable raw manifest. The pinned Hugo package is verified against its published SHA-256 before `sudo dpkg`, and the webhook and alert counts now match the shipped configuration.

Verification:
- Hugo Extended v0.164.0 production build
- no generated root-level internal `href` or `src` values
- all 113 `relref` targets resolved during the build
- actionlint `.github/workflows/docs.yml`
- `make ci` via the pre-push hook

heymrbox:
Addressed in 5c0a375.

The CacheBackend quickstart now shows the complete path to the documented readiness state: configure the subscriber image, wait for the controller rollout, recreate existing engine pods so admission runs again, and send a request to produce the first KV event. The install guide also pins cert-manager to the v1.16.1 version exercised by the repository smoke tests.

PromptTemplate hashing and status fields are now clearly described as planned or reserved behavior. I also rebuilt the JSON-LD metadata from Hugo dictionaries so descriptions and tags are JSON-escaped, replaced the placeholder partial cache key, removed the unused npm manifest, and updated the stale shortcode example.

Verification:
- make ci
- Hugo 0.164.0 production build with the GitHub Pages base URL
- parsed every generated JSON-LD block
- actionlint .github/workflows/docs.yml

heymrbox:
Addressed in 1fd164c.

The CRD reference now states the project’s actual v1alpha1 compatibility policy. The sizing page now keeps distinct prefix keys separate from `(prefix key, replica)` storage entries, corrects the 900K example, and points memory limits at the central `inference-cache-server` Deployment rather than managed CacheBackend workloads.

I also corrected component ownership to match the compiled binaries, documented `/snapshot` and `/policy` as the two directional bridge flows with `/probe` as the functional self-test, and made subscriber injection explicitly opt-in wherever it was described. The doctor page now treats API reachability as prerequisite check 0 followed by nine diagnostics.

Verification:
- make ci
- Hugo 0.164.0 production build
- git diff --check

heymrbox:
Addressed in f3108fc.

The Pages build now has read access to the Pages API, and every GitHub Action in the workflow is pinned to the immutable commit behind its current major tag. Docsy offline search is enabled and the production build emits the project-scoped search index.

The navbar, version selector, tabs, and footer tooltip now use Bootstrap 5 data attributes. LookupRoute documentation no longer claims strict side-effect freedom; it states the actual guarantee that lookups do not mutate cache membership or response-visible routing state while delivered LFU hits may update internal eviction accounting. I also removed the reported trailing whitespace.

Verification:
- make ci
- Hugo 0.164.0 production build
- verified the generated offline search index and project-prefixed references
- actionlint .github/workflows/docs.yml
- git diff --check

heymrbox:
Addressed in bda83b8.

All architectural index-key descriptions now include the `adapter_id` partition. The CacheBackend page treats replica IDs as opaque and explains that backend attribution comes from selectors and injection metadata rather than a pod-name convention.

The Kubernetes-native tenant-isolation example is now a valid Deployment shape with a selector, matching pod labels, and a concrete vLLM image and command. I also removed the extra punctuation from the CachePolicy link.

Verification:
- Hugo 0.164.0 production build
- searched the site for stale four-part index keys and pod-prefix attribution claims
- git diff --check

PR #169, opened 2026-07-28
ci: require docs updates for CRD/proto changes (verify-docs-sync)

EdHasNoLife:
Keeps the user-facing docs from drifting behind the public surface — mirrors the existing proto→`grpc-contract.md` rule, extended to the docs site.

**Rule:** a change to CRD API types (`api/v1alpha1/*_types.go`) or the gRPC contract (`proto/`) must also update the docs (`site/` or `docs/`) in the same PR.

**What is added**
- `hack/verify-docs-sync.sh` + `make verify-docs-sync` — diff-based check (base = `origin/main` locally, PR base in CI).
- `.github/workflows/ci.yml` — enforced on PRs inside the **Lint, Generated Code, Proto** job (a required check). Reuses the base ref already fetched for `buf breaking`.
- PR template + CONTRIBUTING — document the rule.

**Escape hatch:** add the `no-docs-needed` label to a PR to waive it for a genuinely doc-exempt change.

**Scope** is intentionally narrow (near-zero false positives): only the two public-contract surfaces trigger it. Widen `TRIGGER_RE` in the script to cover more surfaces (e.g. the CLI) if desired.

Verified locally: CRD-only change → fails; CRD+docs → passes; proto-only (pre-commit bypassed) → fails; non-surface change → passes.


EdHasNoLife:
Thanks for the review. Both should-fix items flag that the docs-sync gate is coarse — any `docs/` or `site/` change satisfies an api/proto change, rather than the surface-specific doc. That coarseness is intentional:

- **proto → `docs/design/grpc-contract.md` is already enforced precisely** by the pre-commit hook (`.githooks/pre-commit` blocks a `.proto` change that does not also touch that doc). A proto change therefore cannot land without updating the contract doc through normal commits; this CI gate is an additional, coarser net on top of that specific rule — not the only guard.
- **A per-CRD → specific-page mapping would need a hardcoded surface→doc table** that drifts as the docs site is reorganized, producing false failures and ongoing maintenance. The scope is deliberately kept narrow for near-zero false positives.
- The gate is intended as a **reminder** that a public surface changed and the docs should be reviewed; reviewers confirm the *right* page was updated.

Leaving the gate coarse by design. `TRIGGER_RE` / the doc-path check is a one-line change if we later want per-surface precision.

heymrbox (approved):

PR #170, opened 2026-07-30
Refactor the CacheBackend API to separate engine cache, remote storage, and provider ownership

heymrbox:
## Summary

- Separate inference runtime and engine-cache configuration from remote-storage provider selection and ownership.
- Add independent provider adapters and structured remote bindings while retaining legacy compatibility.
- Make canonical SGLang + LMCache host-only when `remoteStorage` is omitted, avoiding implicit Redis provisioning.
- Update admission, generated CRDs, samples, design documentation, and install-smoke coverage.

## Linked issues

CAC-257

### Quality
- [x] CI is green.


heymrbox:
Thanks for catching this. Fixed in d215a62.

The hierarchy discriminator now considers only `runtime`, `lmCache`, and `remoteStorage`, so typed observation remains independent and cannot turn a legacy managed backend into host-only mode. I added a controller regression test proving an observation-enabled legacy backend retains its Deployment, Service, and endpoint. Conflicting legacy and typed model IDs are still rejected at admission.

I also updated the `spec.type` API documentation and regenerated the CRD so the served schema describes engine-cache identity and the independent `remoteStorage` contract correctly.

Verification passed:
- focused API, admission, controller, and pod-webhook tests
- all sample admission checks (21 passed, 2 intentionally skipped)
- full pre-push `make ci` gate

heymrbox:
Addressed in 7c32a74.

Malformed canonical external endpoints now report `spec.remoteStorage.endpoint` in the Ready-condition message, while legacy External resources continue to report `spec.endpoint`. Added regression coverage for both field paths.

Verification:
- `go test ./internal/controller`
- full pre-push `make ci` gate

heymrbox:
Addressed in 8ee3e9c.

Canonical `remoteStorage.endpoint` now uses the same cross-namespace opt-in enforcement as the deprecated endpoint field. Admission also rejects `remoteStorage` in events-only mode and rejects autoscaling on canonical host-only backends, since neither configuration has a provider workload to manage.

I updated the CacheBackend API documentation for the canonical endpoint and observation fields and added regression tests for each validation path.

Verification: `go test ./internal/webhook/v1alpha1` and the full pre-push `make ci` gate passed.

heymrbox:
Addressed in 7b8a0b0.

`lmCache.hostMemory.capacity` now has CRD CEL and webhook validation requiring a positive quantity. The GiB conversion now uses overflow-safe quotient/remainder ceiling arithmetic, shared by the vLLM and SGLang paths, with coverage through `math.MaxInt64`.

Verification:
- `go test ./api/v1alpha1 ./internal/webhook/v1alpha1 ./pkg/adapters/runtime/internal/enginewire ./pkg/adapters/runtime`
- webhook envtest against Kubernetes 1.31 assets
- `make manifests generate`
- full pre-push `make ci` gate

heymrbox:
Addressed in d10597f.

Host-only injection now removes any stale `LMCACHE_REMOTE_URL` and enables the default 20 GiB local CPU tier when `hostMemory` is omitted. Host-only reconciliation also evaluates the connector’s kernel-health and engine-compatibility diagnostics instead of clearing them; events-only still clears both because it loads no connector.

Managed LMCacheServer and Mooncake command overrides now require a non-empty command with non-blank entries in both the CRD schema and admission validation.

Verification:
- focused webhook, controller, engine-wire, and runtime tests
- Kubernetes 1.31 webhook envtest
- `make manifests generate`
- `make verify-samples` (21 passed, 2 intentional skips)
- full pre-push `make ci` gate

heymrbox:
Addressed in f449b45.

Canonical External endpoints now use one provider-aware validator across admission, status reconciliation, and pod injection. Bare `host:port` remains portable; LMCacheServer accepts `lm://`, Mooncake accepts `mooncakestore://`, and Redis accepts only bare `host:port` with a numeric port. Wrong-provider and nested schemes are rejected, while the deprecated External form keeps its LMCache behavior.

I also updated the API docs and the stale SGLang comment, with regression coverage at the admission, controller, pod-webhook, and runtime layers.

Verification:
- Focused Go tests for the affected webhook, controller, and runtime packages
- `make verify-samples` (21 passed, 2 intentional skips)
- Full pre-push `make ci` gate

heymrbox:
Addressed in 9a385ab.

The `backendConfig` section now says plainly that the map is deprecated and legacy-only. It directs canonical manifests to `lmCache`, typed `remoteStorage` provider blocks, `observation`, and `engineOverrides`, and labels the remaining key tables as compatibility behavior.

I also refreshed the reconciler dispatch comment to describe EventsOnly, host-only, External ownership, and managed Redis/LMCacheServer/Mooncake paths.

Verification: the full pre-push `make ci` gate passed.

heymrbox:
Addressed in c05fbe9.

When a legacy resource adopts canonical fields, the defaulter now carries an existing `integration.firstEventTimeout` into `observation.firstEventTimeout` instead of replacing it with five minutes. A migration regression test covers the custom-timeout case.

I also updated the LMCacheServer, Mooncake, and Redis provider comments to point canonical users to the typed `remoteStorage` fields while retaining the deprecated `backendConfig` notes for legacy resources.

Verification:
- `go test ./internal/webhook/v1alpha1 ./pkg/adapters/runtime ./pkg/adapters/runtime/sglang`
- Full pre-push `make ci` gate

heymrbox:
Fixed the failing envtest in a90f87d.

The test still expected the retired top-level `spec.resources` schema default. It now exercises the canonical API: renderer defaults bound a managed LMCacheServer without persisting deprecated fields, and typed `remoteStorage.lmCacheServer.resources` overrides flow to the container.

Verification:
- Kubernetes 1.31 envtest with race detector: `TestIntegrationCacheBackendResources`
- Full pre-push `make ci` gate

heymrbox:
Addressed in 0fa9176.

Managed Redis, LMCacheServer, and Mooncake workload rendering now lives in `pkg/adapters/backend/provider`, along with provider defaults, effective config helpers, and Redis renderer tests. The provider registry no longer imports `pkg/adapters/runtime`; runtime now owns engine injection and keeps only deprecated compatibility delegates for existing callers.

I also refreshed the stale SGLang, vLLM, controller, and design-doc ownership language.

Verification:
- `go test ./pkg/adapters/backend/provider ./pkg/adapters/runtime/... ./internal/controller`
- `go test ./...`
- Full pre-push `make ci` gate

heymrbox:
Addressed in 418a536.

The CacheBackend design now documents the canonical hierarchy end to end: host-only vLLM omits `LMCACHE_REMOTE_URL`, `remoteStorage` selects provider protocol and ownership, external endpoints use provider-specific validation, and the Pod webhook passes a structured binding to runtime injection.

The subscriber and SGLang reference docs now use `observation.modelID` and `runtime: SGLang`. I also refreshed the Mooncake provider, status/readiness, validation, and persistence sections while retaining clearly labeled legacy compatibility behavior.

Verification:
- `make verify-naming verify-no-internal-refs`
- `git diff --check`
- Full pre-push `make ci` gate

heymrbox:
Addressed the final review nits in 077d53f.

- The generated CacheBackend API documentation now explicitly includes host-only LMCache as an endpoint-free injection path.
- The shared reconcile comments now describe both events-only and host-only serverless transitions.

Verification:
- `make manifests generate`
- `go test ./api/v1alpha1 ./internal/controller`
- Full pre-push `make ci` gate

internal/webhook/v1alpha1/cachebackend_webhook.go
```diff
@@ -783,6 +1060,25 @@ func (v *CacheBackendValidator) checkRuntimeAdapter(cb *cachev1alpha1.CacheBacke
...
+	protocol, err := backendadapter.ProtocolFor(storage)
+	if err != nil {
+		return field.ErrorList{field.Invalid(
+			field.NewPath("spec", "remoteStorage", "provider"),
+			storage.Provider,
+			err.Error(),
+		)}
+	}
+	binding := backendadapter.BindingFor(storage, protocol, "")
+	if bindingAware, ok := adapter.(adapterruntime.RemoteBindingAdapter); ok &&
```
fredericsun: [P1] Fail closed when an adapter does not declare binding support

This condition rejects an unsupported binding only when the selected adapter implements `RemoteBindingAdapter`. An adapter that has not implemented the new capability is silently accepted. The shipping SGLang HiCache adapter is one concrete case: a canonical `runtime: SGLang`, `type: SGLangHiCache` resource with Managed Redis passes admission (I reproduced `ValidateCreate == nil`). The controller then renders the Redis workload because it uses the same opt-in type assertion, while pod injection falls back to the legacy endpoint API and HiCache ignores that endpoint. The result can advertise/provision a remote tier that is not in the engine data path.

For canonical resources, please require an explicit `RemoteBindingAdapter` contract instead of treating a missing implementation as compatible. HiCache can implement it as nil-binding-only and reject any non-nil remote storage. Preserve the legacy endpoint fallback only for legacy resources, and add admission/controller/pod regressions for this combination.

fredericsun (commented):
Two actionable issues found while validating the refactor against CAC-257.

[P2] Migrate `inferencecache doctor` to the separated storage model

`pkg/cli/doctor/checks/cachebackend.go` is unchanged by this PR and still classifies a backend with `cb.Spec.Type != CacheBackendTypeExternal`, then requires `status.endpoint` for every backend. Under the canonical API, an external backend has `type: LMCache` and `remoteStorage.ownership: External`, so doctor incorrectly runs the managed-only pod/index checks. Conversely, a valid host-only backend intentionally has no remote storage and an empty endpoint, but doctor reports CB005.

I reproduced both cases with focused tests:
- canonical host-only returned `[CB005]`;
- canonical External returned `[CB002 CB003]`.

Please derive the health model from `EffectiveRemoteStorage()`: skip endpoint checks when storage is nil, and treat `ownership: External` as external while retaining the legacy mapping. Add regressions for both canonical shapes.

fredericsun:
[P2] Organize adapters along one axis and centralize built-in registration

The new API separates runtime/cache wiring from remote-storage provider selection, but the package layout still mixes those concerns and uses inconsistent classification axes:

- `pkg/adapters/runtime` contains the runtime contracts and registry, concrete vLLM implementations, subscriber/kernel-check helpers, and legacy provider shims.
- SGLang and External implementations live in child packages (`runtime/sglang`, `runtime/external`), while vLLM implementations live in the parent contract package.
- `runtime/provider_compat.go` and the legacy `ResolveCacheServer` methods keep runtime packages depending on concrete provider rendering.
- Canonical `External` is a storage ownership value, but `runtime/external` continues to present it as a runtime-adapter category.

This asymmetry already forces `runtime.DefaultRegistry()` to be partial: it cannot import its child implementation packages without a cycle, so controller startup, admission fallback, pod-webhook fallback, and some tests manually register different additions. That creates a concrete registry-drift risk where admission, reconciliation, and pod injection can support different adapter sets while the code still compiles.

Please keep each package on one responsibility axis and add a single composition root. For example:

```text
pkg/adapters/
├── runtime/
│   ├── adapter.go
│   ├── registry.go
│   ├── vllm/
│   └── sglang/
├── storage/
│   ├── binding.go
│   ├── provider.go
│   ├── registry.go
│   ├── redis/
│   ├── lmcacheserver/
│   └── mooncake/
├── observation/
├── compat/
└── builtin/
    └── registry.go
```

`builtin` should be the only place that imports all concrete implementations and constructs the runtime/provider registries; controller, admission, and pod injection should receive that same composition. Legacy External/Mooncake/provider wrappers can be isolated under `compat` rather than shaping the canonical packages.

For this PR, I would at least align the newly refactored runtime/storage/builtin/compat boundaries and add a registry-parity test. The pre-existing `pkg/adapters/engine` (actually the KV-event subscriber) and `engineclient` naming can move in a focused follow-up to keep this already-large PR bounded.

EdHasNoLife:
## Review — direction is right; main gap is an incomplete docs/samples sweep

Read the full diff. The decomposition (orthogonal `runtime` / `type` / `remoteStorage{provider,ownership}`, host-only, typed config blocks, the `pkg/adapters/backend` provider boundary) is the correct shape, backward-compat is handled deliberately, and unit + envtest + install-smoke coverage is strong. No blocking correctness defects. Two threads: an incomplete docs/samples sweep (the actionable one), and a few smaller code notes.

### 📄 Docs / samples sweep is incomplete

Adding `runtime`/`lmCache`/`remoteStorage`/`observation`, removing the `integration.engine` + `resources` kubebuilder defaults, and deprecating `backendConfig` triggers a full sample + doc sweep. This PR updated the primary recipes (`cachebackend-lmcache`, `cachebackend-sglang`, + new host-only) and the core design docs, but leaves the rest on the legacy shape:

- **Samples still legacy — 8 of 10.** The flagship two are `config/samples/cachebackend-mooncake.yaml` (`type: Mooncake` + `backendConfig`) and `config/samples/cachebackend-external.yaml` (`type: External` + top-level `endpoint`) — these are exactly the anti-patterns Examples 1 & 2 of the ticket exist to remove, so the canonical Mooncake/External recipes now teach the deprecated shape. Also untouched: `cache_v1alpha1_cachebackend.yaml`, `cachebackend-events-only`, `-cpu-override`, `-lmcache-cpu`, `-with-engine`, `-with-override`.
- **Docs still on the old surface.** `README.md` still describes `spec.integration.engine` as the runtime identity/default; `docs/quickstart.md`'s getting-started YAML uses `integration.engine` + `backendConfig`; and most importantly `docs/design/kvevent-subscriber-wiring.md` states `--model-id ← spec.backendConfig.model (single source)` — this PR moves the single source to `spec.observation.modelID`, so that design doc now **contradicts the code**. Plus `docs/concepts/cachebackend-engine-binding.md`, `docs/concepts/cachebackend-engine-overrides.md`, `docs/observability/alerts.md`, `docs/reference-stack/README.md`.

None of these break (legacy compat keeps them valid), so it's teaching-the-deprecated-pattern drift, not a bug. Suggest either migrating them in this PR, or explicitly marking the untouched ones as legacy-compat examples + a tracked follow-up. At minimum I'd not ship `cachebackend-mooncake.yaml`, `cachebackend-external.yaml`, and `kvevent-subscriber-wiring.md` as-is, since those directly contradict the new model.

### 🔎 Code notes (non-blocking)

- **Latent nil-deref.** In `internal/webhook/v1alpha1/cachebackend_webhook.go`, `checkRuntimeAdapter` reads `storage.Provider` in the binding-rejection branch, which is reachable with `storage == nil` (host-only) if a binding-aware adapter ever returns `false` for a nil binding. Not triggerable today (both shipping adapters accept nil), but a future host-only-incapable adapter would panic the validating webhook. `dispatch` handles the same case without dereferencing — worth a matching `storage != nil` guard here.
- **Memory-bound default moved schema → webhook.** The 4Gi/8Gi OOM guard is now stamped only by the defaulter for legacy resources, and the legacy renderer path (`effectiveProviderResources` → `cache.Spec.Resources`) has no independent fallback, whereas the canonical path falls back to `defaultCanonicalProviderResources()`. Mitigated by the mutating webhook's `failurePolicy: Fail`, so the guarantee holds on the normal path — flagging as a weakened invariant, not a bug. A legacy fallback in `effectiveProviderResources` would restore defense-in-depth.
- **Silent no-op.** Canonical host-only + `integration.mode: EventsOnly` is admitted and treated as events-only, so a configured `spec.lmCache` host tier is never injected (dispatch checks events-only before host-only; `rejectEventsOnlyMisconfiguration` only forbids `remoteStorage != nil`). Consider rejecting the combo or documenting precedence.
- **Dead const.** `modelBackendConfigKey` in `pkg/adapters/runtime/lmcache_shared.go` has no remaining referent after the subscriber switched to `EffectiveObservationModelID()` (which reads the literal `"model"` key).
- **Test-matrix gap (minor).** Canonical *Managed Mooncake* reconcile isn't covered directly (Redis + LMCacheServer are); the dispatch→`Render`→`reconcileManaged` path is shared, so low risk — a one-line reconcile test would close the matrix.

EdHasNoLife:
## Follow-up — migration-experience notes (non-blocking) + two clean verifications

Went a layer deeper after the first pass. Two items about the **migration experience** (which matters here since a cleaner API is the whole point of this PR), one additive-behavior note, and two things I checked that are fine.

### Worth raising

- **`spec.runtime` is a case-sensitive enum, but the field it replaces was case-insensitive.** `runtime` is `+kubebuilder:validation:Enum=VLLM;SGLang`, so `runtime: vllm` (lowercase) is rejected by the apiserver. The legacy `integration.engine` accepted `vllm`/`vLLM`/`VLLM` interchangeably (lowercased at read time). An operator with `engine: vllm` muscle memory who writes `runtime: vllm` gets a hard rejection. Samples use the correct casing so it's latent — but it's an ergonomic sharp edge in the migration. Consider documenting it, or relaxing the enum to accept common casings.

- **No in-place migration path for existing CRs, and it isn't documented.** Legacy CRs keep working via the compat layer (good), but modernizing one can't be done piecemeal: adding `runtime` to a CR that still carries `backendConfig` flips it to canonical, and `validateCanonicalCacheHierarchy` then rejects `backendConfig`/`resources`. That all-or-nothing behavior is *correct* (no half-migration), but combined with the stale legacy samples there's no worked "convert your old Mooncake/External CR to the new shape" example anywhere, and I don't see a `ValidateUpdate` test for that specific legacy→canonical transition. A short migration section in the design doc + one converted example (and ideally an update-path test) would close it.

- **New Ready reason `HostOnlyActive`.** Additive, but any dashboard/alert keying on `CacheBackend` Ready reasons should be aware it now exists.

### Checked and fine (so the record is complete)

- **Primary design doc is correct.** Confirmed `docs/design/cachebackend-api.md` *removes* the old "CRD-stamped default" / UPDATE-rollout assertions and replaces them with accurate "legacy webhook default … canonical providers apply the same bounded default in their renderer" language. The doc drift I flagged earlier is confined to the peripheral docs (`kvevent-subscriber-wiring.md`, README, quickstart, concepts, alerts).
- **RBAC is clean; hostMemory validation is belt-and-suspenders.** No `config/rbac`/`PROJECT` changes needed — the provider renders the same Deployment/Service/HPA kinds the reconciler already had perms for. And `hostMemory.capacity > 0` is enforced by *both* a CEL `x-kubernetes-validations` rule and the `rejectNonPositiveHostMemoryCapacity` webhook rule, so it's caught even where the CEL `quantity()` lib isn't available.

Net: nothing blocking. The enum casing and the missing migration guide/example are the two I'd actually act on, since they're the operator-facing seams of exactly the API this PR set out to make cleaner.

internal/webhook/v1alpha1/cachebackend_webhook.go
heymrbox: Fixed in 1227985.

Canonical CacheBackends now require their runtime adapter to implement the structured remote-binding contract. Native SGLang HiCache implements that contract as host-only and rejects every non-nil binding, so Redis can no longer be admitted or provisioned for it. Legacy resources retain the endpoint fallback.

Added coverage at the admission, controller, adapter, pod-webhook, and installed-webhook smoke layers.

heymrbox:
@fredericsun, the doctor migration from your review is fixed in 1227985.

It now classifies backends through `EffectiveRemoteStorage()`: External ownership skips managed engine/index checks, while host-only backends keep those checks and skip endpoint reachability. Both canonical shapes have regression tests.

heymrbox:
@EdHasNoLife, addressed the findings in [your review comment](https://github.com/cachebox-project/inference-cache/pull/170#issuecomment-5182494310) in 1227985.

Public CacheBackend samples and recipes now use the canonical `runtime` / `observation` / `remoteStorage` hierarchy, including Mooncake and External ownership. The README, quickstart, subscriber wiring, binding/override docs, alerts, doctor docs, and reference-stack docs were updated as well.

The smaller code notes are covered too: nil-safe admission errors, the legacy 4Gi/8Gi renderer fallback, documented `EventsOnly` precedence, dead constant removal, and canonical Managed Mooncake coverage.

heymrbox:
@EdHasNoLife, addressed the follow-up notes from [this comment](https://github.com/cachebox-project/inference-cache/pull/170#issuecomment-5182715203) in 932a698.

The served CRD and design doc now call out the case-sensitive `VLLM` / `SGLang` enum values. The design doc includes an atomic legacy-to-canonical migration example and links the converted External and Mooncake samples. A `ValidateUpdate` regression proves partial migrations are rejected while the complete replacement is admitted.

`HostOnlyActive` is now documented in the conditions contract.

heymrbox:
@fredericsun, thanks for the package-layout proposal in [this comment](https://github.com/cachebox-project/inference-cache/pull/170#issuecomment-5154956530). I kept the package/composition-root reorganization out of this review-response set because it changes package ownership and dependency wiring beyond the canonical contract corrections in this PR.

This pass closes the concrete registry/binding correctness gap by requiring every canonical runtime adapter to implement the remote-binding contract. The broader runtime/storage/builtin/compat package reorganization remains follow-up work.

heymrbox:
Addressed the latest [Codex review](https://github.com/cachebox-project/inference-cache/pull/170#issuecomment-5182850431) in 5ee3cc2.

- `Offload` now describes all canonical ownership modes: host-only, external storage, and controller-provisioned managed storage.
- The observation interface identifies its no-sidecar example as the deprecated legacy External adapter.
- The Mooncake section and links now use the canonical “Mooncake provider configuration” name.

Verification: `make pre-pr` and the pre-push `make ci` gate both passed.

heymrbox:
Addressed the latest [Codex review](https://github.com/cachebox-project/inference-cache/pull/170#issuecomment-5182958738) in 95306f4.

Canonical `VLLM` + `LMCache` backends with a Mooncake remote binding now apply the `engineHostNetwork` opt-in and `ClusterFirstWithHostNet` after connector injection succeeds. The pod-webhook tests now exercise the canonical hierarchy, including the endpoint-not-ready fail-open path, and admission errors describe the effective Mooncake provider rather than legacy `spec.type`.

The install smoke now pins its Mooncake fixture to `runtime: VLLM`, `type: LMCache`, and `remoteStorage.provider: Mooncake`. I also fixed the stale canonical sample image rewrites and resource assertions that caused the previous install-smoke failure.

Verification: focused adapter/webhook tests, `sh -n`, `make pre-pr`, and the pre-push `make ci` gate all passed.

heymrbox:
Addressed the latest [Codex review](https://github.com/cachebox-project/inference-cache/pull/170#issuecomment-5183078438) in 93e8d25.

The controller now selects the runtime adapter for every Offload backend and validates externally owned storage with the same provider protocol and remote-binding contract used elsewhere. An admission-bypassing `VLLM` + External Redis object now reconciles unmanaged, clears a stale endpoint/Ready condition, and has a regression test. The nil-registry fallback also registers the legacy External adapter to match production wiring.

I also updated the installed External sample smoke to read the canonical `spec.remoteStorage.endpoint` and expect `Type=LMCache`; that was the next stale legacy assertion in the previous failed smoke run.

Verification: controller tests, `sh -n`, `make pre-pr`, and the pre-push `make ci` gate all passed.

EdHasNoLife (approved):

EdHasNoLife:
## Re-reviewed at `93e8d2520` — all prior findings resolved ✅

Went back through every item from the two earlier comments against the 6 new commits, re-read the changed files, and ran the tests on the PR head. Everything is addressed, and substantively.

**Findings resolved**
- **Nil-deref in `checkRuntimeAdapter`** — nil-guarded `storagePath`/`rejectedValue`; the sibling `rejectEngineHostNetwork…` site was hardened the same way. ✅
- **Legacy renderer memory default** — `effectiveProviderResources` now falls back to `defaultCanonicalProviderResources()` when `spec.resources` is nil, restoring defense-in-depth on the legacy path. ✅
- **Host-only + EventsOnly dropping `lmCache`** — precedence documented in the field godoc, CRD, and migration doc. ✅
- **Dead const `modelBackendConfigKey`** — removed. ✅
- **Canonical Managed-Mooncake reconcile coverage** — `TestReconcileCanonicalManagedMooncake` added (plus canonical external/hicache reconcile tests). ✅
- **Samples sweep** — every recipe migrated to the canonical shape, including the two flagships (`cachebackend-mooncake.yaml`, `cachebackend-external.yaml`). ✅
- **Doc drift** — `kvevent-subscriber-wiring.md` now correctly sources `--model-id` from `spec.observation.modelID` (legacy `backendConfig.model` fallback); README/quickstart/concepts/alerts/doctor updated. ✅
- **`runtime` enum case-sensitivity + migration** — documented in godoc/CRD and a new "Migrating a legacy resource" section (+ the pre-launch carve-out governance note). ✅

**Nice catches in the same pass (not from the review)**
- *Fix canonical Mooncake engine networking* — the `engineHostNetwork` guard keyed on `spec.type==Mooncake`, which is never true in canonical shape (`type: LMCache` + `remoteStorage.provider: Mooncake`), so it would have wrongly rejected the flag a Mooncake data plane actually needs. Re-keyed on the effective provider and the hostNetwork/DNS injection extracted into a shared helper. 👍
- *Validate external storage bindings* — a canonical External backend with an unsupported binding now stays unmanaged instead of falsely going Ready.

**Verified locally on PR head:** `go build ./...` clean; `./internal/webhook/...` + `./pkg/adapters/...` green; `./internal/controller -run TestReconcileCanonical` → 8/8 pass (incl. Mooncake).

One non-blocking note for the record: the EventsOnly+`lmCache` precedence and the `runtime` enum casing are resolved by documentation rather than admission enforcement — both fine choices; a hard rejection on EventsOnly+`lmCache` would be the only stricter alternative if you want it surfaced at `kubectl apply`. Not gating on it. LGTM.

PR #172, opened 2026-08-05
CAC-262: Add NFS remote storage for SGLang HiCache

fredericsun:
## Summary

Add NFS as an externally owned remote-storage provider, with initial runtime support for SGLang HiCache.

This PR introduces the following canonical configuration:

```yaml
spec:
  runtime: SGLang
  type: SGLangHiCache
  hiCache:
    ratio: "2.0"
    storagePrefetchPolicy: wait_complete
  remoteStorage:
    provider: NFS
    ownership: External
    nfs:
      server: 192.0.2.10
      path: /hicache
      mountPath: /mnt/hicache
```

NFS is modeled as a generic `remoteStorage.provider`, rather than as an SGLang-specific API. The currently supported runtime/cache pair is `(SGLang, SGLangHiCache)`.

For matching SGLang engine Pods, the webhook injects:

- `--hicache-storage-backend file`
- `--hicache-storage-prefetch-policy <value>`
- `SGLANG_HICACHE_FILE_BACKEND_STORAGE_DIR=<mountPath>`
- An inline NFS volume using the configured server and export path
- A volume mount on the SGLang engine container at `mountPath`

The NFS mount target is externally owned. inference-cache does not create a PV, PVC, StorageClass, NFS export, mount target, Deployment, Service, or HPA.

Because NFS is mounted directly into the engine Pod, it has no CacheBackend network endpoint. The controller acknowledges the current generation but does not publish an endpoint or synthetic `Ready` condition. HiCache readiness remains a separate contract tracked by CAC-260.

This PR does not claim an NFS or SGLang data-plane test. The configured NFS server must be reachable from the engine nodes, and runtime store/read behavior should be verified in the target cluster.

## Change breakdown

1. **`feat: add NFS remote storage for SGLang HiCache`** (`891d949`)

   - Adds `NFS` to the generic remote-storage provider API.
   - Adds typed `remoteStorage.nfs` fields:
     - `server`
     - `path`
     - `mountPath`
   - Adds the typed HiCache `storagePrefetchPolicy` field.
   - Introduces a generic file/NFS backend binding that does not require a network endpoint.
   - Registers External NFS as a provider while limiting runtime consumption to the adapters that explicitly support it.
   - Extends SGLang HiCache validation and injection for file-backed L3 storage.
   - Preserves all-or-nothing and idempotent Pod mutation:
     - Matching existing arguments, environment variables, volumes, and mounts are preserved.
     - Conflicting, malformed, or duplicate reserved configuration causes fail-open admission without partially mutating the Pod.
   - Adds admission validation for:
     - NFS requiring `ownership: External`
     - NFS rejecting `remoteStorage.endpoint`
     - Required and valid server, export path, and mount path
     - `storagePrefetchPolicy` being required with NFS and rejected without NFS
     - Runtime/cache/provider compatibility
   - Handles endpoint-free External bindings without provisioning a managed backend workload or publishing endpoint readiness.
   - Updates doctor classification so NFS is treated as mounted storage rather than a dialable endpoint.
   - Adds a committed NFS sample and updates the CacheBackend design documentation.
   - Adds API, provider, runtime-adapter, admission, Pod-webhook, controller, doctor, sample-admission, and installed-cluster smoke assertions.

## Linked issues

CAC-262

## Validation

Passed locally:

- `make build`
- `make test`
- `make lint`
- `make verify-naming`
- `make manifests generate` — generated CRD and deepcopy code are committed
- `make verify-samples` — 22 passed, 2 intentional skips
- `make ci`
  - Vendor-neutral naming guard
  - Internal issue-reference guard
  - gofmt
  - go vet
  - golangci-lint
  - Prometheus rule validation
  - Documentation synchronization checks
  - Full race-test suite
  - Build
- `git diff --check`

The optional golden-vector verification was skipped by the local CI script because the local Python environment does not have `xxhash` installed.

The installed-cluster smoke script now asserts the NFS lifecycle and server-side dry-run Pod mutation, but the full kind install-smoke was not executed locally for this commit.

### Quality

- [x] `make lint` is clean.
- [x] `make manifests generate` produces no uncommitted generated drift.
- [x] New and changed behavior has unit tests.
- [x] The install-smoke gate contains assertions for the operator-facing CRD, controller, and Pod-webhook behavior.
- [ ] CI is green.

- [x] The CacheBackend design documentation is updated with the NFS contract.
- [x] Backward compatibility was considered for `v1alpha1` consumers; the API changes are additive.
- [x] N/A — `proto/` is unchanged.


EdHasNoLife:
## Review — high quality; no blocking defects, main item is merge coordination with #171

Read the full diff and ran the suite on the PR head (`079ba90`): builds clean, all webhook/adapter/controller/doctor/api packages green. This is a well-structured, thoroughly-tested change. Findings are coordination/awareness-level, not blocking.

### Correctness — strong
- **Layered validation.** `validateCleanAbsolutePath` rejects non-absolute, non-clean (`/hicache/..` → `Clean != value`, good traversal defense), and whitespace; `mountPath` can't be `/`; `ValidateNFSServer` accepts only IPv4/IPv6/DNS-1123; NFS rejects `endpoint`, forces `ownership: External`, and gates `storagePrefetchPolicy` required-with / forbidden-without NFS; the new args + storage-dir env are added to the reserved set. The webhook checks are re-run at the adapter boundary as defense-in-depth against admission bypass — deliberate and correct.
- **Idempotent injection with real conflict detection.** `planHiCacheNFSWiring` matches env by name, volume by reserved name, mount by name-or-path, DeepEqual-compares before add-vs-skip, and rejects conflicts/dups fail-open with no partial mutation. The mount guard correctly catches a foreign mount squatting the target path. Re-injection is a no-op.
- **Endpoint-free binding abstraction is clean:** `ProtocolFile` + `BindingRequiresEndpoint` + `AdapterRequiresEndpointFor` thread consistently through the webhook (`effectiveEndpoint` → `""`), dispatch, and doctor.

### Findings (all non-blocking)

**🟡 Coordinate the merge with #171.** This branch predates #171; both edit the same HiCache dispatch branch, `hicache.go`, `podinjector.go`, and `cachebackend-api.md`, so expect textual conflicts. Substantively: in the merged code NFS-HiCache stays on the endpoint-free-External path → `reconcileUnmanaged`, so it never reaches #171's `reconcileEngineLocal` and gets **no readiness reporting** (consistent with "NFS readiness = CAC-260"). Fine today — but #171's `engineConfigConverged` hardcodes a **nil binding**, so whenever NFS-HiCache is later routed through engine-local readiness, that convergence check will inject without the NFS wiring and mis-Degrade every NFS pod unless it's made binding-aware. Worth a note on CAC-260 so it isn't a surprise.

**🟡 `reconcileUnsupported` broadens the status surface beyond NFS.** Several previously-silent paths (events-only no-adapter, no-runtime-adapter, unsupported external binding, unsupported host-only, no-provider, unsupported managed binding) now publish `Ready=False/Degraded=True` instead of no conditions. Genuine UX improvement and it's tested — but it's a status-behavior change wider than "add NFS," so anything keying on condition presence for those cases should be aware. Additive + v1alpha1 → low risk.

**🟢 Minor — `failOpen` defaults `true` but NFS requires `false`.** A minimal NFS manifest that omits `failOpen` is defaulted to `true`, then rejected; the operator must set `failOpen: false` explicitly. Deliberate (inline NFS mount is a startup dependency), documented, and tested — just an ergonomic wrinkle when copying a host-only HiCache sample. Also note every NFS backend emits the `FailClosedEnabled` Warning event by design.

**🟢 Minor — inline `NFSVolumeSource` gives no mount-option control.** It exposes only `server`/`path`/`readOnly` — no `hard`/`soft`/`timeo`. Under the fail-closed model an NFS outage mid-serving can hang engine I/O (default hard mount) rather than fail-fast. Inherent to inline NFS, not a code issue — worth an operator-facing line in the design doc.

**🟢 Nits.** Two near-identical path validators (`validateCleanAbsolutePath` vs `validateHiCachePath`) could share a helper. `ValidateNFSServer` rejects uppercase hostnames (DNS-1123 strictness) — `NFS.example.com` would be refused; fine, but slightly surprising given DNS is case-insensitive.

### Tests / docs
Excellent lockstep: deepcopy + CRD + schema assertions + validation table + full args/env/volume/mount injection test + binding + provider + doctor branches + controller + committed sample + design doc + install-smoke NFS lifecycle. The PR is upfront that no real NFS/SGLang data-plane test runs — appropriate given the mount is externally owned.

### Verdict
Approve-worthy on its own — nothing blocking. Main actionable item is **merge coordination with #171** (overlap + the nil-binding convergence assumption for any future NFS readiness); the `reconcileUnsupported` broadening and the two operational notes are worth a mention but not gating.

fredericsun:
@EdHasNoLife Thanks for the thorough review.

I addressed the two in-scope operational/documentation notes in `925ef56`:

- Qualified the controller and Pod-webhook comments: fail-open is the default/admission-failure posture, while a successfully injected External NFS backend is an explicitly acknowledged fail-closed serving dependency.
- Documented that inline `NFSVolumeSource` exposes no mount-option control, Linux defaults to a hard mount, a mid-serving outage can block HiCache file I/O, and `storagePrefetchPolicy` cannot interrupt an NFS syscall already blocked in a storage thread.

For the #171 coordination point, I’m intentionally leaving #172’s dispatch unchanged: NFS-backed HiCache remains on the endpoint-free External/`reconcileUnmanaged` path. The nil-binding assumption in `engineConfigConverged` needs to become binding-aware when CAC-260 (or the #171 integration) routes NFS-backed HiCache through engine-local readiness; handling it there avoids changing today’s dispatch contract in this PR.

The broader `reconcileUnsupported` status behavior is intentional and covered by the tests already in this branch.

EdHasNoLife (approved):

PR #174, opened 2026-08-05
Reorganize repository boundaries and project structure

heymrbox:
## Summary

- Move controller/server HTTP wire contracts into `internal/controlplaneapi` while retaining compatibility aliases in `pkg/server`.
- Move shared engine-pod metadata into `internal/enginebinding`, removing controller dependencies on webhook implementations.
- Establish `internal/adapters/builtin` as the complete adapter composition root used by production and fallback paths.
- Add architecture guards, package ownership documentation, and a staged protobuf import migration plan.
- Preserve existing runtime behavior and public wire formats.

## Linked issues

CAC-259

### Quality
- [x] CI is green.

- [x] If CRD API types (`api/v1alpha1/*_types.go`) or the proto contract changed, the **documentation is updated to match** — the docs site (`site/`) and/or the design docs (`docs/`). CI enforces this (`make verify-docs-sync`); add the `no-docs-needed` label to waive a genuinely doc-exempt change.


heymrbox:
Addressed the documentation nit in cc446cc.

The migration rule now says each remaining `pkg/` package must document either its owner or its supported external-consumer contract, which matches the classifications and package docs in this PR.

Verification: the documentation checks passed, and the pre-push `make ci` gate was green.

heymrbox:
Addressed both follow-up findings in 4af4e62.

The CacheTenant webhook now reads the reserved probe tenant ID from `internal/controlplaneapi`, and its admission tests use that same shared constant. This removes the duplicated server-owned literal.

The controller boundary test now walks `internal/controller/` recursively, so production packages added in subdirectories are covered too.

Verification: focused webhook and boundary tests passed, and the pre-push `make ci` gate was green.

heymrbox:
Cleaned up the stale registry terminology in bba6710.

Production and nil-fallback documentation now names the complete built-in composition, package-level comments use `NewCoreRegistry` where appropriate, and the misleading test failure message was updated. I also swept the related controller, webhook, and adapter comments so the old composition description does not linger elsewhere.

Verification: all affected packages passed, and the pre-push `make ci` gate was green.

heymrbox:
Addressed the two final nits in 0b84580.

The CacheTenant admission diagnostic now points to `internal/controlplaneapi` as the probe contract owner, and the recursive controller boundary test no longer repeats the `_test.go` exclusion inside its parse loop.

Verification: both focused packages passed, and the pre-push `make ci` gate was green.

byteforge-go (approved):

PR #175, opened 2026-08-05
CI: Harden runtime images with Distroless checks

heymrbox:
## Summary

- Pin the controller, server, and KV-event subscriber runtime stages to the explicit Distroless Debian 13 non-root image.
- Add automated checks for runtime base, user, entrypoint, payload, shells, and package managers.
- Run source-level and built-image verification in CI.
- Document the container-image hardening policy and verification commands.

## Linked issues

CAC-218

### Quality
- [x] CI is green.

- [ ] If CRD API types (`api/v1alpha1/*_types.go`) or the proto contract changed, the **documentation is updated to match** — the docs site (`site/`) and/or the design docs (`docs/`). CI enforces this (`make verify-docs-sync`); add the `no-docs-needed` label to waive a genuinely doc-exempt change.


heymrbox:
Addressed in 6773626.

The image verifier now requires each expected payload to be exactly one regular executable file, so missing files, directories, symlinks, and non-executable files all fail validation.

I also expanded the forbidden runtime tools to cover additional common shells and package managers, including ash, dash, BusyBox, microdnf, and rpm. The test suite now exercises these cases along with missing images and unavailable Docker commands.

Verification: `make ci` passed locally, and the pre-push `make ci` gate passed before the commit was pushed.

heymrbox:
Addressed in 5fa7877.

Forbidden shells and package managers are now detected by basename across the entire exported filesystem, so alternate locations such as `/usr/local/bin/sh` and `/usr/local/bin/apt-get` cannot bypass the check.

I added regression cases for both alternate paths. The focused verifier suite and the pre-push `make ci` gate pass.

heymrbox:
Addressed in a39f2e3.

The minimal-image verifier now applies forbidden tool names only to regular files, symlinks, and hardlinks. That allows legitimate metadata directories such as `/etc/dpkg` while continuing to reject shell or package-manager payloads anywhere in the image.

I added regression coverage for the `/etc/dpkg` directory and for an alternate-path shell symlink. The focused verifier suite and the pre-push `make ci` gate both pass.

heymrbox:
Addressed in a7f6503.

The payload check now requires the world-execute bit, which guarantees the declared non-root runtime user can execute the binary regardless of archive ownership. A root-owned mode-0700 payload is covered by a new failing test.

I also expanded the basename denylist with the additional common shells and package managers from the review, including fish, csh/tcsh, PowerShell, zypper, and pacman, with focused regression cases.

The verifier suite and the pre-push make ci gate both pass.

EdHasNoLife (approved):

heymrbox:
Addressed the CI-only tar behavior in dbcbbd1.

GNU tar recursively listed descendants when the verifier queried the `/etc/dpkg` directory by name, so a regular metadata file beneath that directory was misclassified as the directory itself. The verifier now pairs `tar -tf` member names with `tar -tvf` entry types in archive order and checks each exact member once.

The regression fixture now puts a child file under `/etc/dpkg`, matching the real Distroless archive shape. The focused suite and pre-push `make ci` gate pass. Local Docker could not run the real image check because the Colima daemon is stopped, so the fresh GitHub image job is the end-to-end GNU tar verification.

heymrbox:
Addressed in 37fb16c.

All three runtime stages now pin `static-debian13:nonroot` to the current multi-architecture index digest, and the verifier policy plus operations documentation use the same immutable reference.

Payload validation now matches exact archive member names using the paired member/mode listings, so whitespace inside an unrelated filename cannot impersonate `/controller`, `/server`, or `/kvevent-subscriber`. A filename-with-spaces regression test covers that case.

The focused verifier suite and pre-push `make ci` gate both pass.

byteforge-go (approved):

PR #184, opened 2026-08-09
Make LMCache server image configurable

heymrbox:
## Summary

- Add `--lmcache-server-image` as the deployment-level fallback for managed LMCache servers.
- Preserve per-CacheBackend overrides through `spec.remoteStorage.lmCacheServer.image`.
- Move the pinned `v0.4.7` baseline from Go code into the shipped Kustomize configuration.
- Add resolution-order unit tests, operator documentation, and install-smoke coverage.

## Linked issues

CAC-193

### Quality
- [x] Every human-authored commit includes a matching DCO `Signed-off-by:` trailer (`git commit --signoff`).
- [x] `make reuse-lint` passes (SPDX headers and licensing metadata are complete).
- [x] CI is green.

- [ ] If CRD API types (`api/v1alpha1/*_types.go`) or the proto contract changed, the **documentation is updated to match** — the docs site (`site/`) and/or the design docs (`docs/`). CI enforces this (`make verify-docs-sync`); add the `no-docs-needed` label to waive a genuinely doc-exempt change.


fredericsun:
this is an improvement for current vLLM LMCache IP mode only using `lm:// standalone server`. It might be refactored soon after we implement vLLM LMCache MP mode. But the logic makes sense to have a fallback LMCache server/worker image in controller. We can use the same fallback idea in MP mode. Will approve. Thank you!

fredericsun (approved):

PR #185, opened 2026-08-10
subscriber: scrape SGLang metrics via a per-engine profile

frankzhouhr:
## Summary

The stats-path metrics scraper was written against vLLM — it hard-coded `vllm:*` metric names and one extraction strategy. Pointed at an SGLang engine it matched no families, yet returned an **all-zero `ReplicaStats` with no error**: a fabricated "healthy" sample that was marked delivered, silently disabling load-aware routing for SGLang replicas (`pressure_factor` degenerates to 1, and holders never get demoted).

This PR teaches the subscriber to observe SGLang correctly and, everywhere, to **fail loud instead of fabricating zeros**:

- **Per-engine metric profile keyed by `hash_scheme`.** vLLM keeps the counter-delta hit-rate (`prefix_cache_hits/queries`); SGLang reads its native gauges (`num_running_reqs` / `num_queue_reqs` / `token_usage`) and takes `cache_hit_rate` as a **direct 0-1 gauge**. The profile is selected from the subscriber's `--hash-scheme`.
- **Fail loud, not fabricate.** A parsed scrape with none of the profile's families (wrong engine/port/scheme) errors; so does a scrape that recognizes the engine but is **missing either load gauge** (partial exposition, metrics disabled, or a `model_name` filter that excludes them). Both go stale/visible rather than reporting `pressure=0`.
- **Multi-series aggregation.** A data-parallel SGLang node emits one series per scheduler (`dp_rank`/`tp_rank`) plus a per-`priority` breakdown. Load **sums the per-rank totals and excludes the `priority="<int>"` duplicates**; usage and hit-rate take the **max across ranks** (bottleneck / best locality). Backward-compatible with single-series vLLM.
- **Metrics endpoint.** The observation sidecar derives `--engine-metrics-url` from the engine container's `--port` (last occurrence wins), falling back to the per-engine default (vLLM `:8000`, SGLang `:30000`).
- **Enable SGLang metrics.** SGLang defaults `/metrics` **off**, so the managed engine paths (LMCache + HiCache) now inject `--enable-metrics` idempotently, and the SGLang reference deployment sets it explicitly.

An absent hit-rate still reports `0`, which conservatively withholds the replica's `TENANT_HOT` fallback — a documented, hint-safe degradation, not a whole-scrape failure.

**Scope note:** EventsOnly SGLang (where `InjectEngineConfig` is a no-op) is not auto-enabled; an operator must pass `--enable-metrics` themselves, and its absence now surfaces as a loud stale signal rather than silent zeros.

## Linked issues

https://linear.app/cachebox/issue/CAC-266/kv-event-ingest-and-metrics-sglang-correctness-audit-routing-path-ok

- [x] Any cloud-specific integration lives in an isolated, optional adapter — never in core controllers, CRD types, the proto contract, or default config.
- [x] `make verify-naming` passes (`✓ core identity is vendor-neutral`).

### Quality
- [x] Every commit includes a matching DCO `Signed-off-by:` trailer.
- [x] `make reuse-lint` passes (508/508 files compliant).
- [x] `make build` and `make test` pass locally (`go test ./...` — all packages pass, incl. `-race` on subscriber / runtime adapters / pod webhook).
- [x] `make lint` clean (gofmt + go vet) and `make ci-lint` clean (golangci-lint: 0 issues).
- [x] `make manifests generate` produces no drift.
- [x] New/changed behavior has unit + end-to-end tests.
- [ ] Operator-facing: injects `--enable-metrics` into managed SGLang engines and updates the SGLang reference deployment. Covered by adapter unit tests; not asserted by install-smoke (no GPU SGLang in the smoke gate).
- [x] `make verify-no-internal-refs` passes.

- N/A — no CRD (`api/v1alpha1`) or proto (`proto/`) changes; the metric profile lives entirely in the subscriber and adapter wiring.


frankzhouhr:
Thanks for the review — addressed all findings in the latest push.

### Should-fix
- **`metrics_scraper.go` — hit-rate baseline advanced before load-gauge validation.** Real bug. Reordered `Scrape`: the both-load-gauges check now runs **before** `hitRate` (and before any other derivation), so a rejected partial scrape never advances the vLLM counter baseline. Added `TestScraperRejectedScrapeDoesNotPoisonHitRateBaseline` (tick1 primes 25/100 → tick2 partial is rejected → tick3 delta is 0.70, proving the baseline stayed at tick1 rather than being poisoned to tick2's 60/120 → 0.4375).
- **`--enable-metrics` not reserved.** Right — added `SGLangEnableMetricsArg` to both `ReservedArgs` lists (LMCache + HiCache), so `engineOverrides.suppressArgs`/`.args` can no longer silently defeat the managed stats path. `isHiCacheReservedArg` only guards operator overrides (not base/injected args), so injection is unaffected. Updated `TestSGLangReservedArgs` / `TestHiCacheReservedArgs`.
- **Verify SGLang metric families/labels/types.** The names, gauge types, `priority=""` total + `priority="<int>"` breakdown, and the `dp_rank/tp_rank/pp_rank` labels were verified against SGLang's `SchedulerMetricsCollector` (`python/sglang/srt/observability/metrics_collector.py`) — the aggregation (sum totals excluding priority breakdowns; max across ranks for the 0–1 ratios) is built directly from that. The fixture header keeps the "pin against the deployed version before shipping" caveat since the reference stack pins a nightly.

### Nits
- **Doc vs. impl for the metrics port.** Updated `kvevent-subscriber-wiring.md` to describe the actual behavior: the port is derived from the engine's own `--port` (last valid wins), falling back to the per-engine default (vLLM `:8000`, SGLang `:30000`).
- **Malformed field comment (stray tab + embedded `//`).** Fixed — split onto proper comment lines.

Local gate: `go build/test ./...` (incl. `-race`), `go vet`, `golangci-lint` (0 issues), `make reuse-lint`, `make verify-no-internal-refs`, and `default-install-smoke`'s HiCache arg contract all pass.

frankzhouhr:
Second round addressed.

### Nit — fixture mapping described unimplemented paths (correct catch)
Fixed. The mapping now documents only what `Scrape` actually consumes:
- `num_running_reqs` / `num_queue_reqs` → pressure (running + waiting)
- `token_usage` → `cache_memory_bytes` (usage × CacheSizeBytes) — **not** pressure or a `k/(1−k)` barrier
- `cache_hit_rate` → `hit_rate` (direct gauge)

`gen_throughput` and `num_used_tokens` are now listed as "present in a real exposition but NOT consumed today," so the fixture no longer describes an aspirational signal path.

### Should-fix — verify SGLang schema against the pinned release
The names, gauge types, and the `priority=""`-total / `priority="<int>"`-breakdown convention were verified against SGLang's `SchedulerMetricsCollector` (`python/sglang/srt/observability/metrics_collector.py`) — the sum-totals-excluding-priority + max-across-ranks aggregation is derived directly from that source. I tightened the const/fixture comments to say "taken from SchedulerMetricsCollector" rather than "docs, pin before shipping," to remove the "unverified" reading.

Honest residual: the reference stack pins a **nightly** SGLang image, and I can't scrape a live GPU SGLang in this environment, so I can't turn the hand-authored fixture into a captured golden vector here. That capture is a reasonable follow-up gated on GPU validation; the code fails **loud** (stale, not fabricated zeros) if a future SGLang build renames or relabels these, so a drift degrades safely rather than silently. Happy to file that follow-up or gate this PR on it — your call.

frankzhouhr:
Addressed — port discovery now scans the container's full argv.

`enginePortFromContainer` (renamed from `...FromArgs`) now reads `Command` **followed by** `Args`, matching how Kubernetes assembles the engine's argv, so a `--port` placed in `command:` (e.g. `["python","-m","sglang.launch_server","--port","40000"]`) is picked up instead of silently falling back to the default. Last valid occurrence across the combined argv still wins.

Added table cases for `--port` in `Command`, and for a `Command` + `Args` split where the `Args` occurrence (later in argv) wins. Full suite + `golangci-lint` (0 issues) green.

frankzhouhr:
Addressed — non-finite gauge values are now rejected.

`aggregateGauge` and `sumCounter` skip any `NaN`/`±Inf` series (Prometheus permits them). A non-finite value no longer counts as "present" and never enters `pressureFrom`/`clamp01`, so:
- a load gauge that is only ever non-finite reads **absent** → Scrape **fails loud** (stale) instead of delivering `NaN` pressure;
- a `NaN` in one dp-rank is skipped while the finite ranks still aggregate, so a single bad scheduler reading doesn't drop the whole load signal.

Tests: `TestScraperNonFiniteLoadGaugeFailsLoud` (NaN running → error) and `TestScraperNonFiniteRankSkippedInAggregate` (NaN rank excluded, finite rank summed). Full suite + `golangci-lint` (0 issues) green.

frankzhouhr:
All three addressed — these were valid.

- **Unknown non-empty scheme silently → vLLM profile.** `NewMetricsScraper` now logs a `WARN` when a non-empty `--hash-scheme` isn't a known profile (falls back to vLLM). A genuine engine/profile mismatch was already caught by fail-loud (`vllm:*` families absent → stale, not wrong stats delivered), but the fallback itself was silent — now it isn't, so a mistyped scheme is visible. Test: `TestScraperUnknownSchemeLogsFallbackWarning`. (Fully decoupling the metric profile from the opaque index `hash_scheme` — e.g. a dedicated `--metrics-profile` flag — is a reasonable follow-up if the team wants custom schemes to be first-class.)
- **`--engine-cache-size-bytes` help was vLLM-only (`usage_perc`).** Updated to name both `vllm:kv_cache_usage_perc` and `sglang:token_usage`.
- **`--cache-tier` help was vLLM-specific.** Clarified it applies to vLLM's tiered gauges; SGLang exposes a single `token_usage` read under `auto`, and explicit `gpu`/`cpu` yield unknown/zero usage for SGLang. (Also tightened `--max-concurrency-ceiling` help to name both engines' scheduler gauges.)

Full suite + `golangci-lint` (0 issues) green.

frankzhouhr:
Both addressed — including the schema verification, now done against the pinned image.

**Schema verification (should-fix).** I verified the SGLang metric contract against the engine's own source **at the exact image pinned in `docs/reference-stack/VERSIONS.md`** (`lmsysorg/sglang:nightly-dev-cu13-20260711-7de33ce8`, i.e. commit `7de33ce8`). `SchedulerMetricsCollector` at that commit emits, verbatim:
- `sglang:num_running_reqs`, `sglang:num_queue_reqs`, `sglang:cache_hit_rate`, `sglang:token_usage` (all gauges)
- the `priority=""` **total** + `priority="<int>"` **breakdown** convention, with `dp_rank`/`tp_rank` labels

— exactly what the fixture and aggregation assume. I updated the const/fixture comments from "re-pin before shipping" to "verified against SchedulerMetricsCollector at the pinned image; re-verify on a version bump," since the earlier wording implied an open gap that no longer exists. The one thing not done here is scraping the *running* binary (no GPU in CI), but the collector source at the pinned commit is authoritative for what it emits — and any future drift fails loud (stale), not silent.

**Stale package doc (nit).** Updated `cmd/kvevent-subscriber` package comment: it now says "next to a vLLM **or SGLang** engine replica" and notes the metric profile is selected by `--hash-scheme`.

Full suite + `golangci-lint` (0 issues) green.

frankzhouhr:
Addressed — `$(VAR)` port references now resolve (with a documented limit).

`enginePortFromContainer` now resolves a whole-value `$(VAR)` reference against the container's own env via `resolveEnvRef`, mirroring the kubelet: `--port=$(ENGINE_PORT)` with `env: ENGINE_PORT=40000` scrapes `:40000` (works in both `--port=` and `--port ` forms, in `command` or `args`).

Documented + tested limitation: a var backed by `valueFrom` (configMap/secret/fieldRef) or an undefined var can't be resolved at admission, so it falls back to the per-engine default — and a wrong port then fails loud on scrape rather than silently reporting the wrong endpoint. New table cases: resolved (equals + space forms), undefined → fallback, `valueFrom`-only → fallback. Design doc updated with the behavior and the limit.

Full suite + `golangci-lint` (0 issues) green.

frankzhouhr:
Both nits fixed. (1) The design-doc adapter inventory now lists SGLang/HiCache alongside SGLang/LMCache at both spots (it does implement `ObservationSidecar`). (2) Qualified the `SGLangEnableMetricsArg` comment: the managed **LMCache and HiCache** paths inject it (both mutate the engine); **EventsOnly** SGLang gets no engine mutation, so an operator running that mode sets it themselves — and its absence surfaces as a loud stale signal, matching the PR's scope note. Full suite + gates green.

heymrbox (approved):

PR #187, opened 2026-08-13
LMCache MP mode

fredericsun:
## Summary

Migrate LMCache integrations from the legacy IP-based data plane to the typed multiprocess data plane.

This PR:

- Adds the typed LMCache multiprocess API and shared MP server renderer.
- Supports both vLLM and SGLang while preserving their engine-specific configuration surfaces.
- Supports:
  - `PodLocal`: one MP server per engine Pod.
  - `NodeLocal`: one engine-demanded MP server per active node, shared by matching engines on that node.
- Preserves inference-system-owned engine scheduling. NodeLocal servers follow scheduled engines using exact-node affinity.
- Gates engine startup until a healthy, ownership-verified same-node server is available.
- Uses Downward API node addressing and never exposes a load-balanced ClusterIP as a CUDA MP endpoint.
- Adds namespace-unique `inferencecache.io/cache-domain` matching to prevent ambiguous CacheBackend ownership.
- Adds per-engine same-node coverage and desired/ready server counts to CacheBackend status.
- Declares NodeLocal host ports so Kubernetes detects same-node conflicts.
- Adds UID-scoped `--shm-name` values to isolate NodeLocal LMCache shared-memory pools.
- Adds typed idle retention so a node-local server can survive temporary engine restarts and is removed after demand expires.
- Removes the legacy LMCache IP wire, `LMCacheConnectorV1`, `lm://` configuration, and LMCacheServer provider path.
- Updates CRDs, generated code, RBAC, samples, documentation, unit/envtest coverage, and install smoke tests.

The migration was validated locally and on SJC dev with vLLM and SGLang TP=1 workloads, including same-node sharing, multi-node placement, L1 store/reset/retrieve, Redis outage/recovery, host-port conflict, ownership/status coverage, UID-scoped shared-memory isolation, and runtime-specific engine metrics.

## Linked issues

N/A

### Quality

- [x] Every human-authored commit includes a matching DCO `Signed-off-by:` trailer (`git commit --signoff`).
- [x] `make reuse-lint` passes (SPDX headers and licensing metadata are complete).
- [ ] CI is green.

Local `make ci`, including the pre-push CI gate and race tests, passes. The CI checkbox should be completed after the remote workflow finishes.

- [x] If CRD API types (`api/v1alpha1/*_types.go`) or the proto contract changed, the **documentation is updated to match** — the docs site (`site/`) and/or the design docs (`docs/`). CI enforces this (`make verify-docs-sync`); add the `no-docs-needed` label to waive a genuinely doc-exempt change.


fredericsun:
I am not addressing the remaining blockers in this PR for two reasons:

- The `v1alpha1` compatibility findings conflict with the documented migration decision: this is a pre-launch alpha API, and the project owner confirmed there are no external consumers or installed legacy objects, so the roadmap explicitly selects an in-place cleanup rather than preserving the old enum/spec/status/selector contract.
- The NodeLocal UID annotation is routing metadata, not authentication; the "proof-of-injection" wording is too strong. NodeLocal is explicitly scoped to one trusted tenant/node domain. Spoofing requires same-namespace Pod creation and can only trigger the deterministic server defined by the operator-owned CacheBackend, not an attacker-selected image or command. Strong provenance for mutually untrusted Pod creators is separate future hardening, and making the cache webhook fail-closed would violate the intentional serving-availability contract.

Therefore I am not making behavioral changes for these blocking findings in this PR.

EdHasNoLife:
## Two operational risks worth resolving before merge

The core MP mechanism (injection safety, endpoint posture, UID isolation, ownership gate) is genuinely well done and builds/tests clean. Two failure-mode risks stand out — both share one shape: **fail-closed and correct on the happy path, but silent on the sad path.** Neither needs an architectural change.

### 🔴 Risk 1 — NodeLocal `/dev/shm` pool is never reclaimed (node-RAM leak)

NodeLocal servers + engines mount a **hostPath** `/dev/shm/inference-cache/<uid>` (tmpfs = node RAM, `DirectoryOrCreate`). The controller deletes the server Pod on idle-expiry / node-demand loss / backend delete (`cleanupLMCacheNodeLocalServerPods` and the idle path just `Client.Delete`), but **Kubernetes never GCs hostPath contents**, and there is no `preStop`, `lifecycle`, finalizer, or reaper anywhere in the PR (the DaemonSet model that could have owned node cleanup was rejected in D5).

This is confirmed by the roadmap's own SJC validation note:
> "the final run also removed only its two exact UID host directories **after confirming that deleted engine/server Pods had left CUDA, torch, and semaphore files behind**."

i.e. cleanup was done **manually**. Because the shm dir/name are "distinct after CacheBackend delete/recreate," every delete→recreate (config change, GitOps redeploy) mints a new UID and **orphans the old GiB-scale pool forever** on every node that ran it — monotonic node-RAM erosion until reboot, not charged to any surviving pod so it won't surface as a pod OOM.

`emptyDir: memory` isn't an option (cross-Pod CUDA IPC needs the shared host inode), so the fix is a reclaimer:
- **`preStop` best-effort unlink** on the server container — covers the graceful 90% (idle-expiry, delete, rollout).
- **Optional reaper DaemonSet** GCing `/dev/shm/inference-cache/<uid>` dirs whose UID no longer maps to a live CacheBackend — covers node-crash/SIGKILL; a reaper does *not* re-introduce the placement inversion D5 rejected.
- At minimum, ship the operator runbook you already followed by hand.

Related: the adversarial isolation item (roadmap "Validate the tenant boundary with unauthorized open/unlink tests") is still unchecked — the boundary is asserted but not yet tested against a hostile co-tenant.

### 🟡 Risk 2 — Ownership gate blocks engine startup forever, with no Kubernetes-native signal

The gate is a `while True: sleep 2` init container (`lmcache_mp_nodelocal.go`) with **no `activeDeadlineSeconds`** on the engine or gate. Fail-closed is defensible here (the `/dev/shm` L1 pool *is* the KV data plane, not a soft cache) — but a misconfigured server (bad image, crash-loop, starved resources) leaves the engine in `Init:0/N` **indefinitely**, surfacing only as gate stdout. No Event, no terminal state, no bounded backoff.

The sharp edge is the interaction with declared host-ports: if backend B reuses backend A's `hostPort` on a shared node, B's server stays `Pending`, B's engines dial **A's** server, the gate's `instance_id` check correctly rejects it (UID mismatch) — and **B's engines wait forever.** A one-line port misconfig becomes a silent deadlock diagnosable only by reading gate logs and correlating UIDs.

Fixes are observability + bounded failure, not redesign:
- Give the gate a configurable **max-wait**, then `exit(1)` → `Init:Error`/CrashLoopBackOff is a native, alertable signal, and kubelet retries (bounded-but-retrying, not silent-forever).
- Have the gate **distinguish "no server yet" from "a different backend owns the port"** (it already holds the UID mismatch) and log/emit that loudly.
- Surface a gate-blocked engine on status as *not-covered with a reason* (the new per-engine coverage / desired-ready counts are the right place), and emit an Event after N failed polls.

I'd treat **Risk 1 as the one closest to blocking** — it's a confirmed resource leak currently cleaned up by hand. Happy to draft the `preStop`/reaper + gate-deadline changes if useful.

fredericsun:
Thanks @EdHasNoLife. These are great findings.

For risk1, I've already noticed the problem. Actually, not only the the shared memory leak, but the way the current design uses shared memory without proper tenant and network segregation is kinda risky. I did not address it in this PR because it needs a bigger scope of design and I put it as the first priority for future work here: https://github.com/cachebox-project/inference-cache/blob/lmcache-multiprocess-migration/docs/design/lmcache-multiprocess-migration-roadmap.md#:~:text=objects%20and%20their%20tmpfs%20pages%20can%20remain%20until%20explicit%20node%20cleanup%20or%20reboot
WIll fix the gap in a separate PR

 For risk2, I understand your concern but this port conflict failure is not silent. The existing implementation already provides several Kubernetes-native signals:
- The NodeLocal server Pod reports the host-port conflict as `Unschedulable`
- `CacheBackend.status` reports `ConnectorReady=False` with reason `NodeLocalHostPortConflict`
- The gate logs the actual and expected `instance_id` when it reaches a server owned by another backend
The engine intentionally remains in init until the correct server becomes available, preventing it from connecting to the wrong cache pool. Once the port conflict is fixed, the gate retries automatically and the engine can continue starting.

EdHasNoLife (approved):

heymrbox (approved):

PR #190, opened 2026-08-19
Add reproducible ranker calibration harness

heymrbox:
## Summary

- Add a reproducible replay harness that sweeps all five ranker knobs through the production LookupRoute implementation.
- Check in an explicitly synthetic boundary fixture, generated per-knob curves, deterministic replay clock, and CI drift verification.
- Keep existing production defaults unchanged until representative captured data supports a retune.

## Linked issues

N/A - tracked internally; no public GitHub issue.

### Vendor-neutral naming (required - see CONTRIBUTING.md)
- [x] No prohibited vendor identity or domains in core surfaces.
- [x] No cloud-specific integration added outside adapters.
- [x] No prohibited domains or namespaces in samples or defaults.

### Quality
- [x] Every human-authored commit includes a matching DCO Signed-off-by trailer.
- [x] make reuse-lint passes.
- [x] make build and make test pass locally.
- [x] make lint is clean.
- [x] make manifests generate produces no drift.
- [x] New behavior has unit tests; calibration package coverage is 98.9%.
- [x] N/A - no operator-facing API or install surface changed.
- [ ] CI is green. Fresh run is in progress for 789191c.

### Contracts
- [x] Calibration methodology and provenance are documented in the design docs.
- [x] Backward compatibility preserved; production defaults and wire behavior are unchanged.
- [x] N/A - proto was not changed.
- [x] N/A - CRD API types and the proto contract were not changed.


heymrbox:
Addressed the deterministic replay and documentation findings in 95063e3.

- Replay now injects each observation timestamp into the production index clock, and replica timestamps come directly from the trace.
- Added a regression test for a TENANT_HOT candidate one millisecond inside the configured age window.
- Narrowed the documentation claim: CI and tests keep the trace, generated result, and code defaults aligned; prose remains review-driven.

Verification: make pre-pr passed, including race tests, builds, generated-code drift, and sample admission.

The synthetic-trace/defaults concern remains open because it needs a product decision: keep the explicitly provisional defaults selected by the synthetic fixture, or retain the previous production defaults until a sanitized captured C1 trace is available.

heymrbox:
Addressed the patch-coverage gap in 24e2579 by covering trace, sweep, observation, and marshal validation/error paths. The calibration package now reports 98.9% statement coverage locally.

Verification: go test -cover ./internal/index/calibration and the full pre-push make ci gate passed.

heymrbox:
Addressed the blocking synthetic-data/defaults finding in 789191c.

- Restored the existing production defaults: PressureWeight 1.0, TenantHotMinHitRate 0.1, and TenantHotMaxAge 5m.
- Decoupled the synthetic fixture candidate from DefaultRankerConfig.
- Updated the calibration docs to state that the synthetic result verifies the harness only and cannot drive production defaults without representative captured data.

Verification: make pre-pr and make cover-check pass; production default code and reference docs now have no drift from main.

heymrbox:
Addressed the latest review in d30cc3a.

The replay trace now carries independent prefix-freshness and replica-stat timestamps, and replay feeds them into the index as separate observations. That lets calibration model fresh cache locality alongside stale pressure data correctly.

I also tightened the harness around the review nits: millisecond durations reject overflow, a nil injected clock safely keeps the default clock, result files are replaced atomically, and command-level tests cover generate/check/stale/malformed flows.

Verification passed locally:
- make pre-pr
- make cover-check
- pre-push make ci gate

heymrbox:
Addressed the two validation findings in 2c4eed2.

Calibration traces now have to include at least one prefix observation and one tenant-hot observation, so every supported ranking mode contributes evidence to the selected five-knob result. Input validation also rejects negative `ttft_budget_ms` values and negative `matched_tokens`, including when a replica does not report the prefix.

Focused regression tests cover each rejected shape, and the full `make pre-pr` plus pre-push `make ci` gates passed.

heymrbox:
Addressed the replay-state findings in c25c817.

Replay now reconstructs the post-sweep production state at each observation timestamp: prefix and statistics reports at or beyond the configured TTL are excluded before lookup, so stale serving entries cannot create a tenant-hot candidate. Sweep values that would overflow the production ranker’s `float32` knobs are also rejected before conversion.

Added regressions for TTL-expired serving state and oversized pressure/bias values. `make pre-pr`, the calibration drift check, and the pre-push `make ci` gate all pass.

heymrbox:
Addressed the trace-contract findings in 6280a83.

The harness no longer treats an unobserved counterfactual as a miss. Every replica row must explicitly declare `outcome_available: true`; captured traces must measure each candidate under an equivalent cache snapshot, while synthetic traces may define ground truth by construction. Incomplete rows are rejected.

Prefix hashes are now opaque `[]byte` values with base64 JSON encoding and a non-UTF-8 round-trip test. Serving-only entries use a request-derived, length-extending key that cannot equal the requested hash, with a regression for the old collision shape. The sweep also accepts zero for the production TTFT and tenant-hot kill switches.

The design note and fixture were updated, and `make pre-pr`, the drift check, focused coverage, and the pre-push `make ci` gate pass.

fredericsun (approved):

